PR #1, opened 2019-12-09
Add openstack to ansible (transible) converter

sshnaidm:
Add code for converting openstack resources config to Ansible playbooks.


jistr:
Thank you!

PR #3, opened 2019-12-09
Rough draft of workflow on Ansible level

jistr:
This depends on PR #2. If PR #2 merges amended, i'll rebase here.

### DIFF
--- a/doc/design.md
+++ b/doc/design.md
@@ -1,3 +1,52 @@
+High-level development goals
+----------------------------
+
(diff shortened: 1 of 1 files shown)


jistr:
Yes i think we'd distribute the collection via Galaxy, hopefully we can request the os-migrate namespace after we build something we can show.

jistr:
I amended according to the comments and added the point about distribution.

PR #6, opened 2020-01-09
CI basic integration for unit tests and docs improvement

ccamacho:


jistr:
@ccamacho can we merge?

ccamacho:
There are some bit I will like to fix, minor fixes and organize the workflows but I can do it in a follow up patch.

Issue #12, opened 2020-01-13
De-duplicate version information


jistr:
Version is stored in two places:

https://github.com/os-migrate/os-migrate/blob/1b5c8c3baddda8def35035706cab97cce2c0ad84/os_migrate/plugins/module_utils/const.py#L1 

https://github.com/os-migrate/os-migrate/blob/1b5c8c3baddda8def35035706cab97cce2c0ad84/os_migrate/galaxy.yml#L12 

Ideally Python code would load the Galaxy manifest for version. However, the galaxy.yml is not shipped as-is when the collection is packaged, it's processed into MANIFEST.json. So we'd be loading either file which doesn't exist during development or during runtime -- both are suboptimal. Some better solution should be found here.

jistr:
Reviewing #19 made me realize that this does have a clean solution -- the collection already does import Python modules in paths that require the collection is built. So we can just forget about galaxy.yml and load MANIFEST.json always.

PR #19, opened 2020-01-13
Build and publish the collection to Galaxy

ccamacho:
This patch builds on each merge a new version of the collection,
then this build should be published to Galaxy.

### DIFF
--- a/.github/workflows/functional-all.yml
+++ b/.github/workflows/functional-all.yml
@@ -57,6 +57,8 @@ jobs:
     - name: Install devstack
       run: |
         cd devstack
+        # Pin a specific version until we fix it.
+        git checkout 46fd6448978897c9e215fc5e038b3d91daa8db5a
         cat << EOF > local.conf
         [[local|localrc]]
         USE_PYTHON3=True
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -1,0 +1,67 @@
+#!/bin/bash
+
+PARAMS=""
+while (( "$#" )); do
+  case "$1" in
+    -k|--galaxy-key)
+      KARG=$2
+      shift 2
+      ;;
+    --) # end argument parsing
+      shift
+      break
+      ;;
+    -*|--*=) # unsupported flags
+      echo "Error: Unsupported flag $1" >&2
+      echo "Please use ./publish.sh [-k <galaxy key> | --galaxy-key <galaxy key>]" >&2
+      exit 1
+      ;;
+    *) # preserve positional arguments
+      PARAMS="$PARAMS $1"
+      shift
+      ;;
+  esac
+done
+
+#
+# Initial variables
+#
+
+namespace=os_migrate
+name=os_migrate
+all_published_versions=$(curl https://galaxy.ansible.com/api/v2/collections/$namespace/$name/versions/ | jq -r '.results' | jq -c '.[].version')
+current_galaxy_version=$(cat actions/galaxy.yml | shyaml get-value version)
+current_galaxy_namespace=$(cat actions/galaxy.yml | shyaml get-value namespace)
+current_galaxy_name=$(cat actions/galaxy.yml | shyaml get-value name)
+publish="1"
+
+#
+# Check all the current published versions and if the
+# packaged to be created has a different version, then
+# we publish it to Galaxy Ansible
+#
+
+for ver in $all_published_versions
+do
+
+    echo "--"
+    echo "Published: "$ver
+    echo "Built: "$current_galaxy_version
+    echo ""
+    if [[ $ver == \"$current_galaxy_version\" ]]; then
+        echo "The current version $current_galaxy_version is already published"
+        echo "Proceed to update the galaxy.yml file with a newer version"
+        echo "After the version change, when the commit is merged, then the package"
+        echo "will be published automatically."
+        publish="0"
+    fi
+done
+
+if [ "$publish" == "1" ]; then
+    echo 'This version is not published, publishing!...'
+    cd ./actions/
+    mkdir -p releases
+    ansible-galaxy collection build -v --force --output-path releases/
+    ansible-galaxy collection publish \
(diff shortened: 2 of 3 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,38 @@
...
+
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
+
+**Desktop (please complete the following information):**
+ - OS: [e.g. iOS]
+ - Browser [e.g. chrome, safari]
+ - Version [e.g. 22]
+
+**Smartphone (please complete the following information):**
```
rbrady: Why is a smartphone something we're adding to the bug template?

.github/workflows/galaxy-publish-collection.yml
```diff
@@ -0,0 +1,27 @@
+name: galaxy-publish-collection
+on:
+  push:
+    paths:
+    - 'os_migrate/galaxy.yml'
```
jistr: -1 here, i don't think we want to automatically publish on every change to galaxy.yml. More generally, do we want to publish via a GH action at all? Would `./toolbox/run make publish` be better?

.github/workflows/galaxy-publish-collection.yml
```diff
@@ -0,0 +1,27 @@
...
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v1
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        sudo apt update -y
+        sudo apt install ansible jq -y
```
jistr: We should add jq installation to `toolbox/build/build.sh` so that the script works from toolbox container too. It's a generally nice tool to have installed anyway IMO.
https://github.com/os-migrate/os-migrate/blob/1b5c8c3baddda8def35035706cab97cce2c0ad84/toolbox/build/build.sh#L11

.github/workflows/galaxy-publish-collection.yml
```diff
@@ -0,0 +1,27 @@
...
+        python -m pip install --upgrade pip
+        pip install tox shyaml
```
jistr: Similarly as jq installation above, these should also go to venv-requirements.txt.
https://github.com/os-migrate/os-migrate/blob/1b5c8c3baddda8def35035706cab97cce2c0ad84/toolbox/build/venv-requirements.txt#L4

scripts/publish.sh
```diff
@@ -0,0 +1,67 @@
...
+#
+
+namespace=os_migrate
+name=os_migrate
+all_published_versions=$(curl https://galaxy.ansible.com/api/v2/collections/$namespace/$name/versions/ | jq -r '.results' | jq -c '.[].version')
+current_galaxy_version=$(cat actions/galaxy.yml | shyaml get-value version)
+current_galaxy_namespace=$(cat actions/galaxy.yml | shyaml get-value namespace)
+current_galaxy_name=$(cat actions/galaxy.yml | shyaml get-value name)
+publish="1"
+
```
jistr: Because of issue #12 we should also check here that versions defined in galaxy.yml and const.py are the same. Simple grep/awk-based check would suffice i think. Not a blocker, i can post it subsequently myself if you want to land this first.

jistr: Ideally we'd fix #12 but i'm not sure about a sufficiently clean solution...

jistr: Ok i just realized the collection always needs to be built to do anything with it and get the proper Python import paths, Python imports fail otherwise. So #12 does have an acceptable solution and we can fix it, so no need to insert anything here. I'll close this comment thread.

.github/ISSUE_TEMPLATE/bug_report.md
ccamacho: removed

.github/workflows/galaxy-publish-collection.yml
ccamacho: done

.github/workflows/galaxy-publish-collection.yml
ccamacho: removed

ccamacho:
I just fixed the comments, added the dependencies in the build script and I pinned the commit that is not breaking the devstack install.

.github/workflows/functional-all.yml
```diff
@@ -57,6 +57,8 @@ jobs:
     - name: Install devstack
       run: |
         cd devstack
+        # Pin a specific version until we fix it.
+        git checkout 46fd6448978897c9e215fc5e038b3d91daa8db5a
```
jistr: Ack let's do this for now, but i don't think Devstack considers itself broken so this issue will likely persist for us, i'll report a tech debt issue to remove the pin.

scripts/publish.sh
```diff
@@ -0,0 +1,67 @@
...
+#
+# Check all the current published versions and if the
+# packaged to be created has a different version, then
+# we publish it to Galaxy Ansible
+#
+
+for ver in $all_published_versions
+do
+
+    echo "--"
```
jistr: We'll probably want to remove these echos going forward, as $all_published_versions will have many entries in the future. But not a blocker for now.

scripts/publish.sh
```diff
@@ -0,0 +1,67 @@
...
+        echo "will be published automatically."
+        publish="0"
+    fi
+done
+
+if [ "$publish" == "1" ]; then
+    echo 'This version is not published, publishing!...'
+    cd ./actions/
+    mkdir -p releases
+    ansible-galaxy collection build -v --force --output-path releases/
```
jistr: We should use the ./scripts/build.sh script to build the collection. That one will also update the version in const.py.

jistr: Actually i like 'releases/' path but build.sh doesn't use it yet. So let's just address both of those things in a follow up patch, i'll merge this one.

PR #28, opened 2020-01-17
Read version information from collection's manifest

### LINKED ISSUE #12: De-duplicate version information

jistr:
We had 2 sources of version information: MANIFEST.json (compiled from
galaxy.yml) and const.py. Now const.py loads the manifest, so
galaxy.yml is the single source of truth in the codebase for version
information. This prevents const.py from getting out of sync with
galaxy.yml.

Closes #12

### DIFF
--- a/os_migrate/plugins/module_utils/const.py
+++ b/os_migrate/plugins/module_utils/const.py
@@ -1,1 +1,19 @@
-OS_MIGRATE_VERSION = '0.1.0'  # can we get this from metadata somehow?
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+import json
+from os import path
+
+
+class Manifest():
+
+    manifest = None
+
+    def __init__(self):
+        if not self.manifest:
+            with open(path.join(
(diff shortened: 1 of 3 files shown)


jistr:
Ok this doesn't work at all when the module is shipped onto target machine, it doesn't have the manifest around for loading. We need a completely different solution, probably just copying the version from galaxy.yml to const.py when building the collection.

jistr:
Closing, will submit an alternative approach.

PR #31, opened 2020-01-20
Export idempotence

jistr:
Depends on PR #23 and #27 

(See commit msgs for details what this PR does.)

PR #32, opened 2020-01-21
Import idempotence

jistr:
Depends on PRs #23, #27, #31 

For idempotence of imports we largely have all the pieces, all that
needs to be done is to serialize the existing network into our format
and compare if the resulting 'params' are the same as the ones
provided to the import_network module.

Crucial is looking at 'params' only and not 'info', as 'params'
contain the important network parameters that are meant to be
migrated, while 'info' contains non-migratable info, e.g. the
network's UUID.

### DIFF
--- a/os_migrate/plugins/module_utils/network.py
+++ b/os_migrate/plugins/module_utils/network.py
@@ -76,3 +76,9 @@ def network_sdk_params(serialized):
     ])
 
     return sdk_params
+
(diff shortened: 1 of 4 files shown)


jistr:
Marking WIP because it depends on #31 which is WIP too.

jistr:
Rebased.

Issue #35, opened 2020-01-23
Make 'build.sh' use 'releases' dir and make 'publish.sh' use 'build.sh'


jistr:
Description at:

https://github.com/os-migrate/os-migrate/pull/19/files/cac3c806de9a483271047bd2f2a4e7fdcdd8513b#diff-70d708edc34d615de58045d62c1c4c1bR64

ccamacho:
Fix in #36

PR #36, opened 2020-01-23
Make 'build.sh' use 'releases' dir and make 'publish.sh' use 'build.sh'

ccamacho:
Fixes: #35

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,10 +13,11 @@ install: os_migrate-os_migrate-latest.tar.gz
 	if [ -n "${VIRTUAL_ENV:-}" ]; then \
 		source /root/venv/bin/activate; \
 	fi; \
+        cd releases; \
 	ansible-galaxy collection install --force os_migrate-os_migrate-latest.tar.gz
 
 clean:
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -13,18 +13,19 @@ install: os_migrate-os_migrate-latest.tar.gz
 	if [ -n "${VIRTUAL_ENV:-}" ]; then \
 		source /root/venv/bin/activate; \
 	fi; \
+        cd releases/
```
jistr: This will not work, there would have to be ; \ at the end of the line (see the line above). Nit: i'd just alter the path on the line below with releases/ instead of doing `cd` :). So this one needs a tweak, otherwise all LGTM.

releases/.gitignore
```diff
@@ -0,0 +1,5 @@
+# Ignore everything in this directory
```
jistr: We could also just add `/releases/*` to the top-level .gitignore to have all the ignoring info in one place, but not a blocker.

Makefile
ccamacho: Done

PR #39, opened 2020-01-24
Wait until the Devstack fix is merged

### LINKED ISSUE #34: Stop pinning devstack version
We pinned devstack because of #33. We should figure out a way to unpin.

https://github.com/os-migrate/os-migrate/blob/27d16410f2d1eb3f3aec37bf6b815e75a2fdd458/.github/workflows/functional-all.yml#L60-L61

ccamacho:
We need to wait until https://review.opendev.org/#/c/704136/
is merged to remove the pin.

Fixes #34

### DIFF
--- a/.github/workflows/functional-all.yml
+++ b/.github/workflows/functional-all.yml
@@ -32,15 +32,16 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
       run: |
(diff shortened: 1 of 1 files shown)


ccamacho:
After the fix is merged I'll just update the change to point to the correct repo to unpin the repo.

ccamacho:
Upstream change merged.

PR #44, opened 2020-02-03
Use tags in functional tests instead of separate playbooks

jistr:
To run subset of functional tests, the current pattern was to create a
full playbook which would only include the desired parts.

An alternative pattern might be to use --skip-tags or --tags with the
test_all.yml playbook. This might allow better mix-and-match execution
for particular test environments and developer scenarios.

Due to env vars not being automatically passed into the toolbox
container, currently there are two ways to test this approach e.g. by
skipping the network tests:

    ./toolbox/shell
    FUNC_TEST_ARGS='--skip-tags test_network' make test-func

and

    ./toolbox/run bash -c "FUNC_TEST_ARGS='--skip-tags test_network' make test-func"

The 2nd case perhaps deserves some improvement (list of env vars we
want to detect and pass into toolbox container automatically?) but
that's beyond the scope of this commit.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 SHELL := /bin/bash
 ROOT_DIR := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
 
-FUNC_TEST_PLAYBOOK ?= all
+FUNC_TEST_ARGS ?=
 
 
 # ANSIBLE COLLECTION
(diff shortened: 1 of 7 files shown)


jistr:
Thanks, so i'll work on some user documentation that should cover those questions. I'm removing WIP+idea labels on this PR and if Carlos likes the approach too, we can merge.

PR #49, opened 2020-02-04
Remove Tmit PoC build requirements

jistr:
I thought i was pasting just vagrant-libvirt package list but in fact
pasted full Tmit build package list. We don't need the Rust bits in
OS-Migrate.


jistr:
Make is there, installed on the line above, but probably something had gcc as a dependency before, but now it's missing. Tomorrow i'll check if there is something else in the same boat as gcc.

PR #50, opened 2020-02-05
Make sure toolbox container RPMs are up-to-date when building

jistr:
We install a bunch of (latest) RPMs but until now we didn't update the
others. This could result in a mixture of RPMs that has never been
tested together in Fedora. Let's update packages before installing
additional ones.

### DIFF
--- a/toolbox/build/build.sh
+++ b/toolbox/build/build.sh
@@ -8,6 +8,7 @@ OS_MIGRATE_DIR=$(realpath "$DIR/../..")
 ### PACKAGES ###
 
 dnf clean all
+dnf -y update
 dnf -y install ansible gcc make python3-devel python3-openstackclient jq shyaml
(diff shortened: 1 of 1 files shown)


toolbox/build/build.sh
```diff
@@ -8,6 +8,7 @@ OS_MIGRATE_DIR=$(realpath "$DIR/../..")
 ### PACKAGES ###
 
 dnf clean all
+dnf -y update
```
ccamacho: Definetely is a good idea to update the container packages before anything else, thanks!

ccamacho:
The image source "fedoraproject" sometimes fails * I was hitting it also.

jistr:
Yea i'm wondering if we should just switch to dockerhub or give it a bit more time to solve the downtime woes this week.

PR #51, opened 2020-02-07
Fix checker of resource sameness, add tests for it

jistr:
When two resources have the same type but their names are undefined,
we should not consider them the same. This is now the case after a
copypasta typo is fixed, and tests are added to verify correct
behavior.


jistr:
Rebased after a merge conflict.

PR #55, opened 2020-02-10
Adapt conventions based on Security Groups discussion

jistr:
Work on Security Groups spawned a discussion about some
conventions. This patch performs following adaptation:

* Use '_info' instead of 'info' for resource properties which are
  irrelevant for importing and idempotence checks. This should further
  reduce a chance of conflict with some real property and allow us to
  ignore any '_info' *also in nested dicts* when importing / checking
  idempotence.

* Better namespacing for Network type
  serialization. 'openstack.network.Network' rather than
  'openstack.network'.


ccamacho:
Awesome!

PR #57, opened 2020-02-11
Advanced name filtering (exact match, regex)

### LINKED ISSUE #22: Filter exported resources by name
We should add resource name filtering, to allow some resources to be worked with and others to be left alone.

In theory this can be done in the *transform* phase between import/export, and could be done manually or automated by the operator according to their needs. But there are reasons to include this basic "processing" it in our export tooling directly:

- This is likely to be a very common use case.

- Filtering during-export rather than after-export means that some resources will not get exported at all, which will make the export process faster. This could make a huge difference especially if/when we happen to add some resources which actually contain binary blobs too.

(We should reiterate that our goal is not to add any features for the *transform* phase (operating on the YAML exports), but name-based filtering during export probably makes sense.)

I can imagine operators might want to filter resources in 2 ways: regexp based (e.g. all networks where name starts/ends with something), list based (export only exact name matches in a list). While the latter can be achieved using the former, it would not be comfortable. So i'd suggest filtering by a list of regexes that have an implicit "logical or" operation applied (matching one regex in the list is enough to consider the resource wanted for export).

jistr:
(see commit msgs)

Closes #22.

### DIFF
--- a/os_migrate/plugins/filter/stringfilter.py
+++ b/os_migrate/plugins/filter/stringfilter.py
@@ -1,0 +1,49 @@
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+from pprint import pformat
+import re
+
+from ansible import errors
+
+
+def stringfilter(strings, queries):
+    """Filter a `strings` list of strigs according to a list of
+    `queries`. Values from `strings` are kept if they match at least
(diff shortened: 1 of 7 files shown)


ccamacho:
Im needing this LGTM

PR #67, opened 2020-02-18
Do not build container with Vagrant in CI

### LINKED ISSUE #24: Allow building toolbox image without Vagrant deps
We want Vagrant included in toolbox image by default so that contributors can simply follow the docs and run functional tests with Vagrant easily. However, we build the image also for CI purposes, and there we don't need Vagrant deps installed, it just prolongs the job. Buildah/docker support --build-arg for build customization, we could use that.

jistr:
We don't need Vagrant in CI and it pulls a lot of deps, skipping its
installation in CI should speed up the jobs.

Closes #24

### DIFF
--- a/.github/workflows/functional-all.yml
+++ b/.github/workflows/functional-all.yml
@@ -95,7 +95,7 @@ jobs:
         openstack --os-cloud devstack-admin --os-region RegionOne router list
     - name: Make the container image to launch the tests
       run: |
-        make toolbox-build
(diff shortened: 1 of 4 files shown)


jistr:
Indeed there is some speed up. Looking at my most recent patches, the container build w/o this patch took 5:06 and 5:55, on this patch the container build took 3:41 and 4:19.

ccamacho:
This is awesome!!! from 18 ->13 mins!

PR #75, opened 2020-02-21
Adds Subnet Export

rbrady:
This patch adds the ability to export subnets from an openstack
cloud into a yaml file.


rbrady:
This patch still needs tests.  Hoping to push that update soon.

ccamacho:
Can you squash all the commits into a single one?

os_migrate/plugins/module_utils/subnet.py
```diff
@@ -0,0 +1,61 @@
...
+        'network_id',
+        'prefix_length',
+        'project_id',
+        'revision_number',
+        'segment_id',
+        'service_types',
+        'subnet_pool_id',
+        'tags',
+        'updated_at',
+    ])
```
jistr: Re params + info, i think we should:

* Remove `location` completely

* Probably move `service_types`  and `tags` to params

* We will need to eventually add `network_name`, `subnet_pool_name` and `segment_name` into params, but it doesn't need to be in this patch. We could instead report a tech debt issue like we did for Sec Groups. And perhaps add a FIXME comment too but if we have an issue it's not necessary i think.

tests/func/run/subnet.yml
```diff
@@ -0,0 +1,21 @@
+# FIXME: Find a way we can test the whole playbook rather than just
```
jistr: Not sure how much you've looked through the func test structure yet, so just a heads up: this `run` file will need `seed` and `cleanup` counterparts.

rbrady:
I just finished adding the unit tests (passing!) and squashing the commits into one and I just now saw the comments left by @jistr.  I'll try to incorporate that feedback over the weekend or on Monday morning.

os_migrate/plugins/module_utils/subnet.py
rbrady: removed location, moved service_types and tags to params and created #88.

os_migrate/plugins/module_utils/subnet.py
```diff
@@ -0,0 +1,60 @@
...
+        'service_types',
+        'tags',
+        'use_default_subnet_pool'
+    ])
+
+    set_ser_params_same_name(info, sdk_subnet, [
+        'created_at',
+        'id',
+        'network_id',
+        'prefix_length',
```
jistr: I'm wondering if prefix_length also belongs to params, i forgot to mention it earlier. But it is only useful with subnet pools which we don't support yet, so no need to keep iterating on this, not a blocker.

tests/func/run/subnet.yml
```diff
@@ -0,0 +1,21 @@
...
+# experience.
+- include_role:
+    name: os_migrate.os_migrate.export_subnets
+  vars:
+    export_subnets_name_filter:
+      - regex: '^osm_'
+
+- name: load exported data
+  set_fact:
+    resources: "{{ (lookup('file', os_migrate_data_dir + '/subnets.yml') | from_yaml)
```
jistr: I think the tests fail because the "resources" name here clashes with the one in network test, they overwrite each other. Sometimes overwriting doesn't matter and saves memory, but in this case we reuse the fact in idempotency testing, so we will have to start namespacing here. `network_resources` and `subnet_resources`

rbrady: thanks.  changing the resources to subnet_resources made the test pass.

Issue #82, opened 2020-02-21
Use id-based iteration for Security Groups and Security Group Rules


jistr:
If we iterate by name and have duplicate namings, the playbook wouldn't behave correctly. See #59.

PR #89, opened 2020-02-25
Use id-based iteration for Security Groups and Security Group Rules

ccamacho:
Closes: #82


jistr:
Actually doesn't need a rebase :)

Issue #80, opened 2020-02-21
Add functional tests for Security Group export


jistr:
We deliberately postponed this from https://github.com/os-migrate/os-migrate/pull/53

PR #91, opened 2020-02-26
Include functional tests from exporting security groups and security group rules

ccamacho:
Include functional tests from exporting security groups and security group rules

Closes: #80
Closes: #81

### DIFF
--- a/tests/func/clean/all.yml
+++ b/tests/func/clean/all.yml
@@ -9,3 +9,15 @@
     apply:
       tags:
         - test_subnet
+
+- include_tasks:
+    file: security_group_rule.yml
+    apply:
+      tags:
+        - test_security_group_rule
+
+- include_tasks:
+    file: security_group.yml
+    apply:
+      tags:
+        - test_security_group
--- a/tests/func/clean/security_group_rule.yml
+++ b/tests/func/clean/security_group_rule.yml
@@ -1,0 +1,8 @@
+- name: remove osm_security_group_rule
+  os_security_group_rule:
+    cloud: "{{ item }}"
+    security_group: osm_security_group
+    state: absent
+  loop:
+    - "{{ os_migrate_src_cloud }}"
+    - "{{ os_migrate_dst_cloud }}"
--- a/tests/func/run/security_group.yml
+++ b/tests/func/run/security_group.yml
@@ -1,0 +1,21 @@
+# FIXME: Find a way we can test the whole playbook rather than just
+# the role. Either make sure we're not in a play here and use
+# import_playbook, or spawn an ansible-playbook subprocess? The latter
+# might be actually a more precise way to test the real end-user
+# experience.
+- include_role:
+    name: os_migrate.os_migrate.export_security_groups
+  vars:
+    export_security_groups_name_filter:
+      - regex: '^osm_'
+
+- name: load exported data
+  set_fact:
+    security_group_resources: "{{ (lookup('file', os_migrate_data_dir + '/security_groups.yml') | from_yaml)
+                   ['resources'] }}"
+
+- name: verify data contents
+  assert:
+    that:
+      - (security_group_resources | json_query("[?params.name == 'osm_security_group'].params.description")
+        == ['OSM security group'])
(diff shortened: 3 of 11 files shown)


os_migrate/plugins/module_utils/network.py
```diff
@@ -239,7 +239,10 @@ def serialize_security_group(sdk_sec):
     """
     expected_type = openstack.network.v2.security_group.SecurityGroup
     if type(sdk_sec) != expected_type:
-        raise exc.UnexpectedResourceType(expected_type, type(sdk_sec))
```
ccamacho: To be discussed

jistr: I found the problem. We should not use methods directly on conn, that's old API:

https://github.com/os-migrate/os-migrate/blob/5fb9018574e082e48b80d523ef009b1d32febef2/os_migrate/plugins/modules/export_security_group.py#L80

We have to use methods under `conn.network`, in this case probably  `find_security_group` with `ignore_missing=False`.

ccamacho: ack, so the old api call is returning a Munch object instead of the one we need (SecurityGroup).

tests/func/clean/all.yml
```diff
@@ -9,3 +9,15 @@
...
+    file: security_group.yml
+    apply:
+      tags:
+        - test_security_group
+
+- include_tasks:
+    file: security_group_rule.yml
+    apply:
+      tags:
+        - test_security_group_rule
```
jistr: I would expect these in inverse order, i.e. first clean up rules then groups, but if it works either way then i guess it's ok. Maybe that will change if we fix the security group rule below.

tests/func/clean/security_group_rule.yml
```diff
@@ -0,0 +1,8 @@
+- name: remove osm_security_group_rule
+  os_security_group_rule:
+    cloud: "{{ item }}"
+    security_group: default
```
jistr: We should leave the default group unchanged and instead use our osm_security_group here.

tests/func/seed/security_group.yml
```diff
@@ -0,0 +1,5 @@
+- os_security_group:
+    cloud: "{{ os_migrate_src_cloud }}"
+    state: present
+    name: default
```
jistr: osm_security_group

tests/func/seed/security_group_rule.yml
```diff
@@ -0,0 +1,7 @@
+- os_security_group_rule:
+    cloud: "{{ os_migrate_src_cloud }}"
+    security_group: default
```
jistr: osm_security_group

tests/func/run/security_group.yml
```diff
@@ -0,0 +1,18 @@
...
+
+- name: load exported data
+  set_fact:
+    security_group_resources: "{{ (lookup('file', os_migrate_data_dir + '/security_groups.yml') | from_yaml)
+                   ['resources'] }}"
+
+- name: verify data contents
+  assert:
+    that:
+      - (security_group_resources | json_query("[?params.name == 'default'].params.description")
```
jistr: osm_security_group

tests/func/run/security_group_rule.yml
```diff
@@ -0,0 +1,18 @@
...
+
+- name: load exported data
+  set_fact:
+    security_group_rule_resources: "{{ (lookup('file', os_migrate_data_dir + '/security_group_rules.yml') | from_yaml)
+                   ['resources'] }}"
+
+- name: verify data contents
+  assert:
+    that:
+      - (security_group_rule_resources | json_query("[?params.security_group_name == 'default'].params.direction")
```
jistr: osm_security_group

tests/func/clean/all.yml
ccamacho: done

tests/func/clean/security_group_rule.yml
ccamacho: done

tests/func/seed/security_group.yml
ccamacho: done

tests/func/seed/security_group_rule.yml
ccamacho: done

tests/func/run/security_group.yml
ccamacho: done

tests/func/run/security_group_rule.yml
ccamacho: done

ccamacho:
All tests passing now using the new API calls.

Issue #83, opened 2020-02-24
Refactoring opportunities somewhere? DRY?


jistr:
We talked earlier about the possibility of refactoring after the basic M1 milestone resources are implemented. Things to look out for:

* Can we DRY something while keeping sanity and generic reusability?

* Should we add some OOP into the mix. One challenge of having resources as classes and e.g. a `Resource` base class is that we need to operate with resources on Ansible level too, not just Python. So if we use custom objects instead of dicts, i wonder if it will work well in places like https://github.com/os-migrate/os-migrate/blob/42afbf3ab38fede427e880d56f21c072d3cf91ec/os_migrate/roles/import_networks/tasks/main.yml#L2-L4 . Something to be investigated...

jistr:
Apparently the supported variable types in Ansible are "String, Numbers, Float, List, Dictionary, and Boolean" [1], i don't think we can just use arbitrary classes. So we'd have to marshal in/out of dictionary whenever crossing the Ansible-Python border. That may defeat the purpose of adopting some OOP for the resources...

[1] https://subscription.packtpub.com/book/networking_and_servers/9781783550630/1/ch01lvl1sec13/variables-and-their-types

PR #94, opened 2020-02-27
Idea: Some OOP refactoring

rbrady:
The purpose of this patch was to demonstrate an area I think OOP
could be saving us time in creating resources.  I focused solely
on the export for now and only the non-nested resources.  I thought
it prudent to just do enough of a patch to get the point across, but
not too much in case we just throw it away.

IMO the main savings is reducing the amount of boilerplate to create
an export.  The SubnetResource and NetworkResource classes illustrate
how a little bit of OOP can help us just focus on adding enough code
to describe the types, info and parameters and not need to add the rest.

Partially addresses #83

### DIFF
--- a/os_migrate/plugins/modules/export_security_group.py
+++ b/os_migrate/plugins/modules/export_security_group.py
@@ -77,7 +77,7 @@ def run_module():
     )
 
     conn = openstack.connect(cloud=module.params['cloud'])
-    sdk_sec = conn.get_security_group(module.params['name'])
+    sdk_sec = conn.network.find_security_group(module.params['name'], ignoring_missing=False)
 
     ser_sec = network.serialize_security_group(sdk_sec)
 
--- a/os_migrate/plugins/modules/export_security_group_rules.py
+++ b/os_migrate/plugins/modules/export_security_group_rules.py
@@ -77,7 +77,7 @@ def run_module():
     )
 
     conn = openstack.connect(cloud=module.params['cloud'])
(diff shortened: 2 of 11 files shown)


jistr:
I started the review but didn't get to publish it before the pull request got closed, i'll just post it subsequently here:

## The Resource class as implemented here



I don't think this is a way forward... With Resource and subclasses we will have to:

* Pass it through into Ansible and back.

* Read it from a file.

For both of those reasons, we will have to stick to simple data types inside a Resource, we shouldn't store a reference to a SDK type at all because we will not be able to deserialize it. I think it might boil down to having a single property self.data or split into self.params self.info and self.type. I was thinking about some design in this respect, maybe if i get some time i would post what i just described, at least how i'd envision the interface + some internals of the Resource class so that it is generic enough and deserializable.

There may be interesting challenges once we get to "weird" resources like Routers and Router Interfaces, it might be challenging to have them fit the mold (depends how restrictive we make the Resource interface) but i don't have any specific things to bring up there right now, unfortunately.

## The const.py refactoring

+1 i think it's nice.

rbrady:
Pass it through to ansible and back?  The interface to ansible is still the module and that doesn’t change.  Maybe you can expand more on this because I don’t see the issue.  All this did was encapsulate some methods and cut down on boilerplate.  It still returns a dictionary just as the previous method implementation did.

Read it from a file.  If you’re talking about this resource class not supporting import yet, that was addressed in the commit message.  If this comment meant something different, please expand.

PR #95, opened 2020-02-27
Import security groups and rules

ccamacho:
Closes: #18 
Closes: #25

### DIFF
--- a/os_migrate/plugins/modules/import_security_group_rule.py
+++ b/os_migrate/plugins/modules/import_security_group_rule.py
@@ -1,0 +1,87 @@
+#!/usr/bin/python
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+ANSIBLE_METADATA = {
+    'metadata_version': '1.1',
+    'status': ['preview'],
+    'supported_by': 'community'
+}
+
+DOCUMENTATION = '''
+---
+module: os_migrate.os_migrate.import_security_group_rule
+short_description: Import OpenStack security group rule
+version_added: "2.9"
+description:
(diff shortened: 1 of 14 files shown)


os_migrate/plugins/modules/import_security_group_rule.py
```diff
@@ -0,0 +1,88 @@
...
+    ser_secrule = module.params['data']
+    secrule_refs = network.security_group_rule_refs_from_ser(conn, ser_secrule)
+    secrule_params = network.security_group_rule_sdk_params(ser_secrule, secrule_refs)
+
+    # The create security group rule method will return an exeption
+    # If it's already created.
+    try:
+        conn.network.create_security_group_rule(**secrule_params)
+    except openstack.exceptions.ConflictException:
+        print("This security group rule already exists")
```
jistr: This is not a super nice approach, i think the cleanest might be: fetch sec group, serialize its rules, see if some of the rules match the one we are creating, skip it if yes. But what you've done here might be quite practical and working. Especially relying on ConflictException and not generic exception (e.g. failing to contact the cloud) makes it ok i think.

jistr: I don't think we should print() though. Rather i'd put this into `result['msg']`. See: https://docs.ansible.com/ansible/latest/reference_appendices/common_return_values.html#msg

os_migrate/plugins/modules/import_security_group_rule.py
```diff
@@ -0,0 +1,88 @@
...
+
+    result['changed'] = True
```
jistr: We should move this at the end of the `try` block. If the rule exists, we don't want to report a change.

PR #97, opened 2020-02-28
registry.fedoraproject.org consistently fails on pulling images

ccamacho:
We had several CI issues related to errors when pulling the
container image from registry.fedoraproject.org

Maybe we wont hit this anymore if we use Docker Hub.
Also, the image shuold be exactly the same.


jistr:
We need full image url but in general +1 on the change, we've had enough of the failures to switch the registry i think.

PR #98, opened 2020-02-28
Idea: Some OOP refactoring

rbrady:
The purpose of this patch was to demonstrate an area I think OOP
could be saving us time in creating resources.  I focused solely
on the export for now and only the non-nested resources.  I thought
it prudent to just do enough of a patch to get the point across, but
not too much in case we just throw it away.

IMO the main savings is reducing the amount of boilerplate to create
an export.  The SubnetResource and NetworkResource classes illustrate
how a little bit of OOP can help us just focus on adding enough code
to describe the types, info and parameters and not need to add the rest.

### DIFF
--- a/os_migrate/plugins/module_utils/const.py
+++ b/os_migrate/plugins/module_utils/const.py
@@ -1,3 +1,8 @@
+from __future__ import (absolute_import, division, print_function)
+from enum import Enum
+__metaclass__ = type
+
+
 OS_MIGRATE_VERSION = '0.1.0'  # updated by build.sh
 
 # Main serialization sections
@@ -10,3 +15,21 @@ RES_TYPE_NETWORK = 'openstack.network.Network'
 RES_TYPE_SECURITYGROUPRULE = 'openstack.network.SecurityGroupRule'
 RES_TYPE_SECURITYGROUP = 'openstack.network.SecurityGroup'
 RES_TYPE_SUBNET = 'openstack.subnet.Subnet'
+
+
+# NOTE: OOP
+# enums to replace the const values above.  Still provides . access to values
+# but easier to read IMO.  The tradeoff that I see is when you want access to
+# the string value under the type, you have to call .value
+# const.RES_TYPE_NETWORK vs ResourceType.NETWORK
+class ResourceType(Enum):
+    NETWORK = 'openstack.network.Network'
+    SECURITY_GROUP_RULE = 'openstack.network.SecurityGroupRule'
+    SECURITY_GROUP = 'openstack.network.SecurityGroup'
+    SUBNET = 'openstack.subnet.Subnet'
+
+
+class Sections(Enum):
+    PARAMS = 'params'
+    TYPE = 'type'
+    INFO = '_info'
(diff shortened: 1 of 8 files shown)


os_migrate/plugins/module_utils/serialization.py
```diff
@@ -116,3 +121,65 @@ def _recursive_trim(obj):
             return obj
 
     return _recursive_trim(resource)
+
+
+class Resource:
```
jistr: As discussed on IRC, i think this approach would give us trouble on import, i will propose an idea PR as well to put my IRC words into some code to read :)

os_migrate/plugins/module_utils/const.py
```diff
@@ -10,3 +15,21 @@
...
+    NETWORK = 'openstack.network.Network'
+    SECURITY_GROUP_RULE = 'openstack.network.SecurityGroupRule'
+    SECURITY_GROUP = 'openstack.network.SecurityGroup'
+    SUBNET = 'openstack.subnet.Subnet'
+
+
+class Sections(Enum):
+    PARAMS = 'params'
+    TYPE = 'type'
+    INFO = '_info'
```
jistr: The "namespaced" constants are nice, +1.

jistr:
Will move the discussion from closed PR here:
https://github.com/os-migrate/os-migrate/pull/94#issuecomment-592543695

Ryan wrote:

> Pass it through to ansible and back? The interface to ansible is
> still the module and that doesn’t change. Maybe you can expand more
> on this because I don’t see the issue. All this did was encapsulate
> some methods and cut down on boilerplate. It still returns a
> dictionary just as the previous method implementation did.

The problem is that we will want to return something from the
read_resources module (
https://github.com/os-migrate/os-migrate/blob/d361567cd7886e2f54055790be32c5b7b96e142f/os_migrate/plugins/modules/read_resources.py
). Currently it is a list of dicts, what we'd like in an ideal world
is a list of Resource subclass instances, but that's not possible in
Ansible. So our resource class will have to be easily marshalable
from-dict + to-dict, which the class in this PR isn't because it
internally saves references to OpenStack SDK objects, and they cannot
be just loaded from dict easily.

Ryan wrote:

> Read it from a file. If you’re talking about this resource class not
> supporting import yet, that was addressed in the commit message. If
> this comment meant something different, please expand.

I think "we'll address import later" is not a viable approach in this
case. In my understanding, we'd have to rework the approach from this
PR substantially to make the Resource class of use during import. It
is ok to add the code later but IIUC the design here goes in a
direction which pretty much prevents adding import functionality.

Again this boils down to -- we need to restore the Resource class from
a YAML file, we won't be able to just load an OpenStack SDK object
from our resource YAMLs.

The potential alternative of only having the `content` parameter
defined when going from SDK but not when going from a file is IMO not
a good design. We'd essentially have 2 different class hierarchies
munged into one -- we'd have to treat Resource internals differently
when loading from file vs. creating from SDK object, which would be
quite confusing and probably error prone too.

I'm proposing an alternative in #105. I tried to keep your DRY approach to
serialization (marking the patch as co-authored) but at the same time i'm
trying to make the Resource easily go to-dict + from-dict.

rbrady:
This patch is obsoleted by #107 and #115.

Issue #100, opened 2020-03-02
Security group rule export writes just one rule


jistr:
The problem is writing is outside of for loop, so it writes just the last rule iterated over:

https://github.com/os-migrate/os-migrate/blob/265e730d89f9b837b44d773dca15248d9b949290/os_migrate/plugins/modules/export_security_group_rules.py#L82-L92

This intermittently fails func tests in my env.

ccamacho:
This is officially our first bug!!!

PR #101, opened 2020-03-02
Write resource correctly when iterating over resources

ccamacho:
We are writing only the last item in the array to the output file.

Closes:#100


os_migrate/plugins/modules/export_security_group_rules.py
```diff
@@ -88,9 +88,9 @@ def run_module():
         sec_refs = network.security_group_rule_refs_from_sdk(conn, sec_rule_obj)
         ser_sec = network.serialize_security_group_rule(sec_rule_obj, sec_refs)
 
-    result['changed'] = filesystem.write_or_replace_resource(
-        module.params['path'], ser_sec)
+        filesystem.write_or_replace_resource(module.params['path'], ser_sec)
 
+    result['changed'] = True
```
jistr: `result['changed']` should be True only if at least one of the write_or_replace_resource calls returned true

PR #103, opened 2020-03-02
Implement Router export

### LINKED ISSUE #70: Export Routers

jistr:
Note that this doesn't export router's interfaces on private subnets,
they will have to be exported/imported via a separate resource.

Closes #70

### DIFF
--- a/os_migrate/plugins/module_utils/reference.py
+++ b/os_migrate/plugins/module_utils/reference.py
@@ -2,6 +2,54 @@ from __future__ import (absolute_import, division, print_function)
 __metaclass__ = type
 
 
+def network_name(conn, id_, required=True):
+    """Fetch name of Network identified by ID `id_`. Use OpenStack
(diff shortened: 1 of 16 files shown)


os_migrate/plugins/module_utils/reference.py
```diff
@@ -2,6 +2,54 @@
 __metaclass__ = type
 
 
+def network_name(conn, id_, required=True):
```
ccamacho: We might then replace calls like https://github.com/os-migrate/os-migrate/blob/b183976cceb0bc48207aa9214b1cea9dab3e2a1f/os_migrate/plugins/modules/export_subnet.py#L79

jistr: This method only returns name if the ID is non-null, it does not return the whole network. It is to DRY the "if not None, give me <something>" use cases, which otherwise result in boilerplate in the calling code.

PR #105, opened 2020-03-03
WIP Resource class

jistr:
This is a very rough idea just for discussion, the code is not runnable.

### DIFF
--- a/os_migrate/plugins/module_utils/resource.py
+++ b/os_migrate/plugins/module_utils/resource.py
@@ -1,0 +1,236 @@
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+import ansible_collections.os_migrate.os_migrate.plugins.module_utils \
+    as collection_utils
+
+class Resource():
+
+    resource_type = 'UNDEFINED'
+
+    # Took the idea of property lists for serialization from Ryan's
+    # code, will mark the patch co-authored.
+
+    # Serialization
+    params_from_sdk = []
+    params_from_ref = []
+    info_from_sdk = []
+    info_from_refs = []
+
+    # Seserialization
+    sdk_params_from_params = []
+    sdk_params_from_ref = []
+
+    # This method should only be called by when creating subclasses
+    # and Resource should not be instantiated directly, not sure if we
+    # have a way to enforce that in Python.
+    def __init__():
+        self.data = {
+            const.RES_TYPE: None,
+            const.RES_PARAMS: {},
+            const.RES_INFO: {},
+        }
+
+    # This constructor will not have to be overriden by child classes,
+    # will work OOTB.
+    @classmethod
+    def from_data(cls, data):
+        res_type = data.get('type', None)
+        if res_type != self.resource_type:
+            raise UnexpectedResourceType(self.resource_type, res_type)
+
+        obj = cls()
+        # Barring resource type errors, we allow Resource subclass
+        # construction with invalid data so that we don't break on
+        # loading a data file, but we explicitly fail with full list
+        # of errors (rather than a single exception) when all
+        # resources are validated with is_data_valid() and/or
+        # data_errors().
+        obj.data = data
+        # Just in case the data didn't contain those keys (invalid
+        # data), set to empty defaults so that all other methods can
+        # rely on these existing rather than calling dict.get().
+        obj.data.set_default(const.RES_PARAMS, {})
+        obj.data.set_default(const.RES_INFO, {})
+        return obj
+
+    # I'd rather opt for intra-method composition of superclass
+    # private methods than calling the superclass method of the same
+    # name.
+    @classmethod
+    def from_sdk(cls, sdk_resource, conn):
+        raise NotImplemented("This method needs to be provided by a subclass")
+
+    def import_or_update(self, conn):
+        raise NotImplemented("This method needs to be provided by a subclass")
+
+    def is_data_valid(self):
+        len(data_errors()) == 0
+
+    # Common interface for validating the resource definition. If
+    # someone edited the resource by hand and made a wrong YAML
+    # nesting for example, we should be catching it here.
+    def data_errors(self):
+        errors = []
+        # ... here we validate that params and info exist
+
+    # This is a bit of scifi right now but i could see e.g. Router
+    # class being able to validate that the Network/Subnet resoruces
+    # referenced from it already exist in the target cloud.
+    def are_import_prerequisites_met(self, conn):
+        # SCIFI!
+
+    def needs_update(self, target):
+        # Check if `sdk_res` needs to be updated to match the state
+        # defined in `self`. We just need to move `_trim_info` from
+        # serialization.py.
+        current_trimmed = _trim_info(current)
+        target_trimmed = _trim_info(target)
+        return current_trimmed != target_trimmed
+
+    def _data_from_sdk_and_refs(self, sdk_res, refs):
+        # we could move these methods from serialization.py to private
+        # under Resource
+        set_ser_params_same_name(params, sdk_res, self.params_from_sdk)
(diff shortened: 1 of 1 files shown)


jistr:
Kind of "pseudocode" right now but i'd like to discuss the design ideas on this and hopefully push forward after the discussion.

os_migrate/plugins/module_utils/resource.py
```diff
@@ -0,0 +1,236 @@
...
+        # Check if `sdk_res` needs to be updated to match the state
+        # defined in `self`. We just need to move `_trim_info` from
+        # serialization.py.
+        current_trimmed = _trim_info(current)
+        target_trimmed = _trim_info(target)
+        return current_trimmed != target_trimmed
+
+    def _data_from_sdk_and_refs(self, sdk_res, refs):
+        # we could move these methods from serialization.py to private
+        # under Resource
```
jistr: params, info = self.params_and_info()

os_migrate/plugins/module_utils/resource.py
```diff
@@ -0,0 +1,236 @@
...
+        if res_type != self.resource_type:
+            raise UnexpectedResourceType(self.resource_type, res_type)
+
+        obj = cls()
+        # Barring resource type errors, we allow Resource subclass
+        # construction with invalid data so that we don't break on
+        # loading a data file, but we explicitly fail with full list
+        # of errors (rather than a single exception) when all
+        # resources are validated with is_data_valid() and/or
+        # data_errors().
```
jistr: This "load anything, validate explicitly" vs. "validate always on load" is something that could be challenged though. There are different benefits and drawbacks to each.

os_migrate/plugins/module_utils/resource.py
```diff
@@ -0,0 +1,236 @@
...
+        # serialization.py.
+        current_trimmed = _trim_info(current)
+        target_trimmed = _trim_info(target)
+        return current_trimmed != target_trimmed
+
+    def _data_from_sdk_and_refs(self, sdk_res, refs):
+        # we could move these methods from serialization.py to private
+        # under Resource
+        set_ser_params_same_name(params, sdk_res, self.params_from_sdk)
+        set_ser_params_same_name(params, refs, self.params_from_sdk)
```
jistr: type params_from_refs

os_migrate/plugins/module_utils/resource.py
```diff
@@ -0,0 +1,236 @@
...
+
+    # Serialization
+    params_from_sdk = []
+    params_from_ref = []
+    info_from_sdk = []
+    info_from_refs = []
+
+    # Seserialization
+    sdk_params_from_params = []
+    sdk_params_from_ref = []
```
jistr: typos here `refs` rather than `ref`

os_migrate/plugins/module_utils/resource.py
```diff
@@ -0,0 +1,236 @@
...
+        # would be just the same as what we already have in network.py
+
+
+
+
+# EXAMPLE in import_network module which gets 'data' as a parameter:
+net = Network.from_data(data)
+# there would be more if/else around this, checking existing network
+# perhaps but you get the idea
+conn.network.create_network(net.sdk_params(conn))
```
jistr: This was what i thought of initially but we can probably make the interface nicer with a full encapsulation of `import_or_update` as written below.

os_migrate/plugins/module_utils/resource.py
```diff
@@ -0,0 +1,236 @@
...
+        'qos_policy_id',
+    ]
+
+
+    @classmethod
+    def from_sdk(cls, sdk_resource, conn):
+        # >>> Internal composition over inheritance! <<<
+        # This is more future-proof, composable, less restrictive
+        # approach than calling the same method via super.
+        obj = cls()
```
rbrady: This becomes the new bolierplate.  It's going to be the implementation for most if not all.  If it's the common case, why not have this be the default method and override IF you are planning to deviate from it?

os_migrate/plugins/module_utils/resource.py
```diff
@@ -0,0 +1,236 @@
...
+        obj._data_from_sdk_and_refs(sdk_resource, _refs_from_sdk(sdk_resource, conn))
+        # Let's not make special lists for sorted params/info, just sort them?
+        obj._sort_param('availability_zone_hints')
+        obj._sort_info('subnet_ids')
+        return obj
+
+    ## TBD: SOME MORE DRY HERE? Can we refactor something more into reusable bits?
+    def import_or_update(self, conn):
```
rbrady: not material to the review, but everything else in openstacksdk is "create" or "update".  I'd like "create_or_update" here for the method name.

os_migrate/plugins/module_utils/resource.py
```diff
@@ -0,0 +1,236 @@
...
+
+class Resource():
+
+    resource_type = 'UNDEFINED'
+
+    # Took the idea of property lists for serialization from Ryan's
+    # code, will mark the patch co-authored.
+
+    # Serialization
+    params_from_sdk = []
```
rbrady: How often are these param and info names different?  I understand the case where we're doing the name/id mapping, but for the rest of them they are most likely the same always, right?

jistr: Params and info are usually completely non-overlapping. Perhaps i don't understand where the question is headed though -- is there some potential code amendment that you are thinking of?

rbrady: What I mean here is we have "params_from_sdk" and "sdk_params_from_params".  These items are shown in the Network subclass below to be the same.  Since we're controlling the names used on export,  we know what the names will be on import.  If a name changes in openstacksdk, then we'd probably have a os-migrate version that pins to a specific openstacksdk version.  

There isn't a case now where we are using different keys for these values, and if there was, there would need to be custom code in the serialize and deserialize methods anyway to do the mapping between them.  

tl;dr: we're optimizing for a case that doesn't exist now and is unlikely to exist in the future.  If it did happen to appear, the developer could account for this when writing the subclass.  I think this falls under YAGNI.

os_migrate/plugins/module_utils/resource.py
jistr: Done

os_migrate/plugins/module_utils/resource.py
```diff
@@ -0,0 +1,236 @@
...
+
+    def _sort_param(self, param_name):
+        self.info()[param_name] = sorted(self.info()[param_name])
+
+
+# This would obviously go to network.py but i want to keep the "idea
+# pull request" simple.
+class Network(Resource):
+
+    params_from_sdk = [
```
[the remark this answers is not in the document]
jistr: Done `sdk_params_from_params = params_from_sdk`

os_migrate/plugins/module_utils/resource.py
jistr: Done

os_migrate/plugins/module_utils/resource.py
jistr: I see now, yes so for now i reduced the boilerplate into a oneliner `sdk_params_from_params = params_from_sdk`. We are probably going to need something advanced when we get to RouterInterfaces, but it may also be that simple separation won't suffice and we'll have to fully override quite a lot of the Resource base class, so perhaps indeed we could just drop this separation altogether. We can always re-add it if we find the case for it. (The good thing with having it in Resource is that **if** the separation is needed, it allows us to still reuse bits in subclasses rather than reimplementing existing methods and just changing one variable name. But i think i'm starting to agree with the YAGNI point of view.)

jistr:
I will close this PR and create a new draft one, will mark the unresolved discussion point in there.

PR #107, opened 2020-03-04
Resource class

jistr:
We created a Resource class as the base class for all our exported
resources. It improves the DRYness of our overall code and is suitable
for both export and import. It uses method composition internally so
that it can be easily "disconnected from network" for purposes of unit
testing.

The common interfaces are now backwards compatible, e.g. the
add_or_replace method for exporting resources now accepts both direct
resource data or an instance of Resource.

Co-Authored-By: Ryan Brady <rbrady@redhat.com>

Closes: #83

### DIFF
--- a/os_migrate/plugins/module_utils/network.py
+++ b/os_migrate/plugins/module_utils/network.py
@@ -3,54 +3,18 @@ __metaclass__ = type
 
 import openstack
 
-from ansible_collections.os_migrate.os_migrate.plugins.module_utils import const
-from ansible_collections.os_migrate.os_migrate.plugins.module_utils import exc
-from ansible_collections.os_migrate.os_migrate.plugins.module_utils import reference
-from ansible_collections.os_migrate.os_migrate.plugins.module_utils import serialization
+from ansible_collections.os_migrate.os_migrate.plugins.module_utils \
+    import const, exc, reference, resource, serialization
 from ansible_collections.os_migrate.os_migrate.plugins.module_utils.serialization \
     import set_sdk_params_same_name, set_ser_params_same_name
 
 
-def serialize_network(sdk_net, net_refs):
-    """Serialize OpenStack SDK network `sdk_net` into OS-Migrate
-    format. Use `net_refs` for id-to-name mappings.
+class Network(resource.Resource):
 
-    Returns: Dict - OS-Migrate structure for Network
-    """
-    expected_type = openstack.network.v2.network.Network
-    if type(sdk_net) != expected_type:
-        raise exc.UnexpectedResourceType(expected_type, type(sdk_net))
+    resource_type = const.RES_TYPE_NETWORK
+    sdk_class = openstack.network.v2.network.Network
 
-    resource = {}
-    params = {}
-    info = {}
-    resource[const.RES_PARAMS] = params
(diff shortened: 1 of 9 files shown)


os_migrate/plugins/module_utils/resource.py
```diff
@@ -0,0 +1,198 @@
...
+    sdk_class = 'UNDEFINED'
+
+    # Serialization
+    info_from_sdk = []
+    info_from_refs = []
+    params_from_sdk = []
+    params_from_refs = []
+
+    # Deserialization
+    sdk_params_from_params = []
```
jistr: As discussed in #105 this is probably YAGNI and we can use `params_from_sdk` in its place.

os_migrate/plugins/module_utils/resource.py
```diff
@@ -0,0 +1,198 @@
...
+    def _set_sdk_params_same_name(cls, ser_params, sdk_params, param_names):
+        """Copy values from `ser_params` into `sdk_params` for all keys in
+        `param_names` (list of strings), only for values in `ser_params`
+        which aren't None.
+        """
+        for p_name in param_names:
+            cls._set_sdk_param(ser_params, p_name, sdk_params, p_name)
+
+    # Not meant to be overriden in majority of subclasses.
+    @staticmethod
```
jistr: The only difference between static and class methods (line 71, line 81) is whether we need to call another static/class method from within. I'm playing with an idea to make them all `classmethod` for consistency and simply not use the first param in some of them. It probably doesn't matter at all in practice though :).

os_migrate/plugins/module_utils/network.py
```diff
@@ -94,24 +41,48 @@ def network_sdk_params(ser_net, net_refs):
...
+    def from_sdk(cls, conn, sdk_resource):
+        obj = super(Network, cls).from_sdk(conn, sdk_resource)
+        obj._sort_param('availability_zone_hints')
+        obj._sort_info('subnet_ids')
+        return obj
+
+    def create_or_update(self, conn):
+        refs = self._refs_from_ser(conn)
+        sdk_params = self._to_sdk_params(refs)
+        existing = conn.network.find_network(sdk_params['name'])
```
jistr: DRYing this into Resource class can come in a separate commit. We cannot simply make a variable for something like `find_network` here, pointing to a static method reference in OpenStack. We'll need to look up that method dynamically on the run-time `conn` object. So DRYing this logic will be more involved and require using intra-language reflection (looking up methods by name).

We can decide separately if we want to introduce this amount of magic or not for making things DRY, but i think this PR already improves things a lot and is worth merging even without this part done.

os_migrate/tests/unit/test_network.py
```diff
@@ -1,54 +1,141 @@
...
+import openstack
 import unittest
 
-from ansible_collections.os_migrate.os_migrate.tests.unit import fixtures
 from ansible_collections.os_migrate.os_migrate.plugins.module_utils \
-    import network
+    import const, network
+
+
+def sdk_network():
```
jistr: Initially i wondered whether these fixtures will be used from multiple unit tests, but it looks like they will be specific to the Network unit test. So i'm moving them from `fixtures.py` here.

PR #108, opened 2020-03-05
Run Ansible lint

ccamacho:
First patch to run lint checks in the project


toolbox/build/venv-requirements.txt
```diff
@@ -8,3 +8,6 @@ pytest-xdist>=1.30.0
 pycodestyle>=2.5.0
 voluptuous>=0.11.7
 yamllint>=1.20.0
+flake8==3.7.7
+ansible-lint==4.1.0
+bashate
```
jistr: Nit: worth keeping alphabetic order here? Should we specify minimum version for bashate?

Makefile
```diff
@@ -33,6 +33,13 @@ os_migrate-os_migrate-latest.tar.gz:
 
 # TESTS
 
+test-lint:
```
jistr: We should add it to test-fast target like: `test-fast: test-lint test-sanity test-unit`. The idea of it was "all checks which don't need OpenStack running".

.github/workflows/linters.yml
```diff
@@ -0,0 +1,34 @@
+---
+name: linters
```
jistr: I'm wondering if it's worth another job or if the unit test job should simply run the `test-fast` target and test all lints + unit tests. The setup time (building container etc.) is probably comparable or longer to the time of actually executing linters + unit tests? I guess i'd prefer just one job for all "blocking checks which do not require running an OpenStack env, and are therefore fast to execute", but it's not a strong opinion, i'd be fine with separate jobs too in case we collectively feel it's a better way.

ccamacho: Im checking the actual difference between jobs and workflows, and it seems the workflows are totally independent.

What we can use is to define 2 jobs, one for consistency checks and then the functional tests execution which depends on the consistency check passing.

toolbox/build/venv-requirements.txt
ccamacho: yeahp, exporting the requirements usually adds the version and sorted alphabetically.

Makefile
ccamacho: yeahp, done.

scripts/linters.sh
```diff
@@ -0,0 +1,30 @@
...
+#
+# Flake
+# TODO: Have the Ansible and YAML lint fixed first
+#flake8 --exclude releasenotes,.tox --ignore E125,E251,E402,H405,W503,W504
+
+# Bash hate
+#find ./ -not -wholename "*.tox/*" -and -not -wholename "*.test/*" -and -name "*.sh" -print0 | xargs -0 bashate -v --ignore E006
+
+#Yaml lint
+#find ./ -not -wholename "*.tox/*" -and -name "*.yml"  -print0 | xargs -0 yamllint
```
jistr: Ack i assume this will be enabled in another PR, LGTM.

jistr:
I looked at CI logs too, they seem good.

PR #110, opened 2020-03-05
Fix YAML nits

ccamacho:


ccamacho:
Too many changes, still missing some.

tests/func/clean/all.yml
```diff
@@ -1,41 +1,47 @@
-- include_tasks:
-    file: router.yml
+---
+- name: Include router tasks
+  include_tasks: router.yml
+  args:
```
ccamacho: Based on 4th example from: https://docs.ansible.com/ansible/latest/modules/include_tasks_module.html

jistr: Ack this is good.

tests/func/independent/validate_data_dir.yml
```diff
@@ -1,16 +1,19 @@
+---
 - name: validate valid dir
+  changed_when: false
```
jistr: True

tests/func/independent/validate_data_dir.yml
```diff
@@ -1,16 +1,19 @@
...
     OS_MIGRATE="$HOME/.ansible/collections/ansible_collections/os_migrate/os_migrate"
-    OS_MIGRATE_DATA="{{playbook_dir}}/data/validate_data_dir/valid"
+    OS_MIGRATE_DATA="{{ playbook_dir }}/data/validate_data_dir/valid"
     ansible-playbook -v \
         -i $OS_MIGRATE/localhost_inventory.yml \
         -e "os_migrate_data_dir=$OS_MIGRATE_DATA" \
         $OS_MIGRATE/playbooks/validate_data_dir.yml
 
 - name: validate invalid dir
+  changed_when: false
```
jistr: True

tests/func/seed/network.yml
```diff
@@ -7,6 +9,7 @@
 #     description: osm_net test network
 #     state: present
 - name: workaround - create osm_net via CLI
+  changed_when: false
```
jistr: true

jistr:
So we need to change at least the `changed_when` to `true` and i would like to stop enforcing the `---` in yaml, but it's not an absolute blocker.

toolbox/vagrant/ansible/roles/devstack/tasks/start.yml
```diff
@@ -1,4 +1,6 @@
+---
 - name: start devstack
+  changed_when: false
```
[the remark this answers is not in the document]
ccamacho: done

os_migrate/localhost_inventory.yml
```diff
@@ -1,3 +1,4 @@
+---
```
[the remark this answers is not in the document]
ccamacho: Well it's already fixed everywhere and it should be something standard for YAML in the beginning of each file, I dont think it hurts having it as it's already done :)

tests/func/independent/validate_data_dir.yml
ccamacho: done

tests/func/independent/validate_data_dir.yml
ccamacho: done

tests/func/seed/network.yml
ccamacho: done

os_migrate/localhost_inventory.yml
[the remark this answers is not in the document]
jistr: Yea we'll just have to keep doing it in every new file too :grin:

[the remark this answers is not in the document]
jistr: Discussed on IRC, we don't have 100% the same feeling about this but it's a small issue that is easy to reverse if we decide that later. We should merge lints ASAP so that we can continue our work with lints enabled and removed risk of conflicts with the lint patches. So merging.

PR #111, opened 2020-03-05
Python lint checks

ccamacho:


os_migrate/tests/unit/test_resource.py
```diff
@@ -87,6 +87,7 @@ def test_from_data_invalid(self):
         data['type'] = 'some.InvalidType'
         with self.assertRaises(exc.UnexpectedResourceType):
             res = FakeResource.from_data(data)
+            self.assertEqual(res.data, data)
```
jistr: This shouldn't be here. If it complains that res is unused, then we should remove the `res =` above.

PR #112, opened 2020-03-06
Run ansible-lint as the last of the linters

jistr:
Ansible-lint is the slowest linter, running it last should speed up
catching Python and YAML lint issues via `test-fast` target.


ccamacho:
Ill wait for the job to finish and then merge it.

Issue #113, opened 2020-03-06
Unnecessary "---" added to YAML files


rbrady:
In f8a6e882baec3f4ab5ce550a833f51bd811f4295, a sequence of characters, "---", was added
to all of the yaml files in the project.  This code is optional for yaml and denotes the
start of a yaml file[1].  The issue I have with this change is we are adding arbitrary
characters to the code and then enforcing that on future contributions, but only for
aesthetic purposes.  There is no value in adding this sequence.  IMO form should follow
function. "Add only what you need, and then style that appropriately."  

 - We're adding 4 bytes per yaml file.  It won't be a significant impact to storage or transport speeds, for every 256 yaml files we will add 1k to overall project size.
 - The yaml parser does not require this.  We're not enforcing this on files parsed for import[2].
 - There is little chance someone will look at one of our yaml files and not recognize that it's yaml even if it's missing "---" and "..."
 - We're not even consistent with linting for options as we're not linting for "..." to denote the ending of all files.

I respectfully suggest we remove the linting rule for "---" and come to an agreement as a project about adding code or rules to the project should provide some value to either users or developers.

[1] https://docs.ansible.com/ansible/latest/reference_appendices/YAMLSyntax.html
[2] https://github.com/os-migrate/os-migrate/blob/master/os_migrate/plugins/module_utils/filesystem.py#L39

jistr:
I wasn't happy with "---" either but wanted to get the lints landed quickly because they did a lot of small edits not significant to how the project works, but with significant collision risk on any ongoing work.

Removing the "---" specifically, should have minimal collision risk. Personally i'm +1 on removal.

PR #114, opened 2020-03-06
remove start of file indicator

ccamacho:
Closes: #113


ccamacho:
Im gona fix also this warning:

find: warning: -wholename .tox/ will not match anything because it ends with /.

PR #127, opened 2020-03-17
Start the container MySQL service

ccamacho:
Closes: #128

### DIFF
--- a/.github/workflows/consistency-functional.yml
+++ b/.github/workflows/consistency-functional.yml
@@ -63,11 +63,11 @@ jobs:
       # mysql:
       #   image: mysql:5.7
       #   env:
-      #     MYSQL_ROOT_PASSWORD: devstack
+      #     MYSQL_ROOT_PASSWORD: root
       #     MYSQL_ALLOW_EMPTY_PASSWORD: true
       #     MYSQL_DATABASE: devstack
       #   ports:
-      #   - 3306/tcp
+      #     - 3306:3306
       #   options: >-
       #     --health-cmd="mysqladmin ping"
       #     --health-interval=10s
@@ -79,6 +79,11 @@ jobs:
         uses: actions/setup-python@v1
         with:
           python-version: ${{ matrix.python-version }}
+      # FIXME: We should use the MySQL service ASAP
+      - name: Start the MySQL service
+        run: |
+          sudo service mysql restart
+          sudo systemctl status mysql.service
       - name: Install dependencies
         run: |
           # We use ppa:projectatomic repo to install podman
@@ -92,8 +97,12 @@ jobs:
           # installed or not
           sudo apt install python3-httplib2 -y
       - name: Verify MySQL connection from container
-        run: |
-          mysql --host 127.0.0.1 -uroot -proot -e "SHOW DATABASES"
+        run: >-
+          mysql --host 127.0.0.1 \
+                --port 3306 \
+                -uroot \
+                -proot \
+                -e "SHOW DATABASES"
       - name: Make the container image to launch the tests
         run: |
           NO_VAGRANT=1 make toolbox-build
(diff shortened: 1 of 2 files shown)


.github/workflows/consistency-functional.yml
```diff
@@ -126,7 +128,7 @@ jobs:
           RABBIT_USERID=guest
           RABBIT_PASSWORD=guest
           # Pre-requisite
-          ENABLED_SERVICES=mysql,key
```
ccamacho: This worked before as it was running in the same container, in this case is external and we can restart it.

.github/workflows/consistency-functional.yml
```diff
@@ -119,16 +121,17 @@ jobs:
...
           MYSQL_HOST=127.0.0.1
           RABBIT_HOST=127.0.0.1
           RABBIT_USERID=guest
           RABBIT_PASSWORD=guest
           # Pre-requisite
-          ENABLED_SERVICES=mysql,key
+          ENABLED_SERVICES=key
           # Neutron
-          ENABLED_SERVICES+=,q-svc,q-agt,q-dhcp,q-l3,q-meta,neutron
+          # ENABLED_SERVICES+=,q-svc,q-agt,q-dhcp,q-l3,q-meta,neutron
```
ccamacho: oki so if we skip the neutron install the devstack install finish, I need to check why neutron is not starting.

.github/workflows/consistency-functional.yml
```diff
@@ -119,16 +121,17 @@ jobs:
...
           SERVICE_HOST=127.0.0.1
           MYSQL_HOST=127.0.0.1
           RABBIT_HOST=127.0.0.1
           RABBIT_USERID=guest
           RABBIT_PASSWORD=guest
           # Pre-requisite
-          ENABLED_SERVICES=mysql,key
+          ENABLED_SERVICES=key
           # Neutron
-          ENABLED_SERVICES+=,q-svc,q-agt,q-dhcp,q-l3,q-meta,neutron
```
ccamacho: Removing neutron services it finished the install we need to debug why is not working, maybe something related to the DB

ccamacho:
I got from logs:


2020-03-17T18:16:49.4030324Z [00;32mDEBUG neutron.agent.linux.utils [[00;36m-[00;32m] [01;35m[00;32mRunning command (rootwrap daemon): ['ovs-vsctl', '--timeout=5', '--id=@manager', '--', 'create', 'Manager', 'target="ptcp:6640:127.0.0.1"', '--', 'add', 'Open_vSwitch', '.', 'manager_options', '@manager'][00m [00;33m{{(pid=53754) execute_rootwrap_daemon /opt/stack/neutron/neutron/agent/linux/utils.py:103}}[00m[00m



2020-03-17T18:16:49.4024202Z [01;31mERROR ovsdbapp.backend.ovs_idl.idlutils [[00;36m-[01;31m] [01;35m[01;31mUnable to open stream to tcp:127.0.0.1:6640 to retrieve schema: Connection refused[00m[00m

ccamacho:
So, using the in-place mysql service from GitHub actions seems not to work ok, the only difference I can see is that when mysql runs in the same container than devstack we are restarting the mysql service.

Im trying now another approach with is installing and configuring mysql as we had it before.

.github/workflows/consistency-functional.yml
jistr: It still looks like we have Neutron API available. I guess that's the `q-svc` in the list. Not sure what `neutron` stands for but the error with connecting to port 6640 that you pasted earlier is related to OVS, so perhaps OVS is the culprit here. Let's merge to unblock CI.

.github/workflows/consistency-functional.yml
```diff
@@ -79,6 +79,11 @@ jobs:
         uses: actions/setup-python@v1
         with:
           python-version: ${{ matrix.python-version }}
+      # FIXME: We should use the MySQL service ASAP
+      - name: Start the MySQL service
+        run: |
+          sudo service mysql restart
+          sudo systemctl status mysql.service
```
jistr: We should probably use a container going forward? But let's unblock.

PR #133, opened 2020-03-20
Security group idempotence

rbrady:


rbrady:
I updated with the web editor to fix the all.yml file, but it caused a CI failure for "trailing spaces".  I couldn't see any trailing spaces, so I rebased the patch locally and fixed the all.yml file again in an editor and pushed the update.

PR #136, opened 2020-03-25
Fix Subnet Update

rbrady:
During the idempotency testing, the subnet update calls to the API
failed for ip_version and network_id.  These parameters are accepted
for the create, but not the update.  Sorting the allocation_pools and
dns_nameservers migtigated the condistion that caused this error.


jistr:
I'll rebase.

PR #137, opened 2020-03-25
Fixes CI

rbrady:
The CI started failing today when attempting to run the checkout
source step.  I started researching online and found there is a
race condition issue with the actions/checkout@v1[1].  This patch
changes the checkout action to use @v2 which supposedly addresses
this issue.

[1] https://github.com/actions/checkout/issues/23

### DIFF
--- a/.github/workflows/consistency-functional.yml
+++ b/.github/workflows/consistency-functional.yml
@@ -18,7 +18,7 @@ jobs:
       matrix:
         python-version: [3.7]
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v1
         with:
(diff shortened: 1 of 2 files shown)


ccamacho:
CDN looks down.

.github/workflows/consistency-functional.yml
```diff
@@ -71,7 +71,7 @@ jobs:
           --health-timeout=5s
           --health-retries=3
     steps:
-      - uses: actions/checkout@v1
```
ccamacho: We should change both:

https://github.com/os-migrate/os-migrate/pull/137/files#diff-37c7821a4a9822799e5e8cfcd0b9d699R21

ccamacho: Sorry but the link does not seem to work, we have also a call to the checkout@v1 action in line 21

jistr:
I added another commit that fixes the lint target so that it installs the collection for ansible linter to pass. I have no idea why we didn't hit this before, maybe something changed in the ansible linter.

PR #139, opened 2020-03-26
Add Sagi as co-author, remove Transible code

jistr:
Folks who want to read OS-Migrate code and contribute could mistakenly navigate into the openstack_ansible dir and start learning that (unused) code instead of the code we really use. Parts of it inspired OS-Migrate though, so let's add Sagi as a co-author.


jistr:
CC @sshnaidm

sshnaidm:
@jistr thanks! Glad it helped.

Issue #143, opened 2020-03-27
Implement new authentication mechanism


jistr:
Something that would be usable in Universal Conversion Hosts, which our current mechanism isn't. We shouldn't copy user's clouds.yaml into the UCH.

PR #144, opened 2020-03-27
New auth

jistr:
This is a first step towards solving #143. Once we agree on the basics here, we'll implement the new auth for the rest of the resources.


jistr:
@rbrady @ccamacho I basically went for "the other of the two" authentication mechanisms supported by the official OpenStack Ansible modules. I don't think there are other sensible choices. If we do anything else, we'll have to stop using the official modules (e.g. `os_networks_info`, `os_subnets_info` etc.). It is not completely out of question to do that, but i don't see enough benefits there to justify reimplementing that code. The auth + auth_type method should probably work well for us and for Universal Conversion Host too i hope.

Issue #142, opened 2020-03-27
Subnet update fails


jistr:
When subnet needs to be updated, it fails with: "Cannot update read-only attribute ip_version", we probably need to drop that parameter from the update request. There may be more parameters like this.

rbrady:
Closed as a duplicate of #140 .  It has the same error message.

jistr:
There are two different bugs though so we should have 2 different issues -- one is for update_subnet getting triggered at all (which means the idempotence is broken), the other is for update_subnet failing (which means the update is broken). So we should reopen this one i think. In PR #145 you're fixing the update, but not the idempotence.

jistr:
So bug #140 should now be fixed by PR #149, but should the subnet update get triggered by a real change in a real use case, it will still fail. Bug #142 is still not fixed. Reopening.

PR #145, opened 2020-03-27
Fixes Subnet Update

### LINKED ISSUE #140: Subnet import fails on functional tests - subnet import not idempotent
Still failing on subnet import

TASK [import_subnets : import subnets] *****************************************
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: openstack.exceptions.BadRequestException: BadRequestException: 400: Client Error for url: http://127.0.0.1:9696/v2.0/subnets/osm_subnet, Cannot update read-only attribute ip_version
failed: [localhost] (item={'_info': {'created_at': '2020-03-26T15:14:23Z', 'id': '286c5f76-2788-4104-a808-391a1c063797', 'network_id': '8593bc62-3f45-46ad-9aeb-5f581a71d778', 'prefix_length': None, 'project_id': '32f2a33cd400488f993795407f9d9d11', 'revision_number': 0, 'segment_id': None, 'subnet_pool_id': None, 'tags': [], 'updated_at': '2020-03-26T15:14:23Z'}, 'params': {'allocation_pools': [{'end': '192.168.0.254', 'start': '192.168.0.2'}], 'cidr': '192.168.0.0/24', 'description': '', 'dns_nameservers': ['8.8.8.7', '8.8.8.8'], 'gateway_ip': '192.168.0.1', 'host_routes': [{'destination': '0.0.0.0/0', 'nexthop': '12.34.56.78'}, {'destination': '192.168.0.0/24', 'nexthop': '192.168.0.1'}], 'ip_version': 4, 'ipv6_address_mode': None, 'ipv6_ra_mode': None, 'is_dhcp_enabled': True, 'name': 'osm_subnet', 'network_name': 'osm_net', 'segment_name': None, 'service_types': [], 'subnet_pool_name': None, 'use_default_subnet_pool': None}, 'type': 'openstack.subnet.Subnet'}) => changed=false
ansible_loop_var: item
item:
_info:
created_at: '2020-03-26T15:14:23Z'
id: 286c5f76-2788-4104-a808-391a1c063797
network_id: 8593bc62-3f45-46ad-9aeb-5f581a71d778
prefix_length: null
project_id: 32f2a33cd400488f993795407f9d9d11
revision_number: 0
segment_id: null
subnet_pool_id: null
tags: []
updated_at: '2020-03-26T15:14:23Z'
params:

rbrady:
Intermittently, the subnet import functional tests fail for ip_version
or network_id attributes.  This patch removes the ip_version and
network_id from the update args.

closes #140

### DIFF
--- a/os_migrate/plugins/module_utils/resource.py
+++ b/os_migrate/plugins/module_utils/resource.py
@@ -190,8 +190,8 @@ class Resource():
         existing = self._find_sdk_res(conn, sdk_params['name'])
         if existing:
             if self._needs_update(self.from_sdk(conn, existing)):
-                update_params = self._remove_readonly_params(sdk_params)
-                self._update_sdk_res(conn, update_params['name'], update_params)
+                self._remove_readonly_params(sdk_params)
+                self._update_sdk_res(conn, sdk_params['name'], sdk_params)
                 return True
         else:
             self._create_sdk_res(conn, sdk_params)
@@ -257,7 +257,6 @@ class Resource():
         for name in self.readonly_sdk_params:
             if name in sdk_params:
                 sdk_params.pop(name)
-        return sdk_params
 
     # Used when creating params for SDK calls, should be overriden in
     # majority of child classes.
--- a/os_migrate/tests/unit/test_resource.py
+++ b/os_migrate/tests/unit/test_resource.py
@@ -128,8 +128,8 @@ class TestResource(unittest.TestCase):
 
     def test_remove_readonly_param(self):
         res1 = FakeResource.from_data(valid_fakeresource_data())
-        sdk_params = res1._remove_readonly_params(valid_fakeresource_data())
-        self.assertFalse('readonly_param' in sdk_params)
+        res1._remove_readonly_params(res1.params())
+        self.assertFalse('readonly_param' in res1.params())
 
     def test_create_and_update_all_ok(self):
         res = FakeResource.from_data(valid_fakeresource_data())


os_migrate/plugins/module_utils/subnet.py
```diff
@@ -70,6 +70,11 @@ def _find_sdk_res(conn, name_or_id):
 
     @staticmethod
     def _update_sdk_res(conn, name_or_id, sdk_params):
+        # intermittently the subnet update fails when trying to update with a
+        # message that states the ip_version or network_id attributes cannot
+        # be updated.  See #140 for more information.
+        sdk_params.pop("ip_version")
+        sdk_params.pop("network_id")
```
jistr: Now it complains about `cidr`. Also we should fix the commit message -- this doesn't fix #140 (idempotence), it fixes #142 (making update work). Either one of them fixed is enough to unblock CI, but they are distinct bugs.

rbrady: IMO we have a third issue here.  If I can run the func tests for the PR locally and they pass with no issues, but then fail in CI, then local tests become worthless.

rbrady: These should be the only fields not for update in Neutron for Subnets: https://github.com/openstack/neutron/blob/8c7ca776d8cbe967a8bbe773ab38c361414a7068/neutron/objects/subnet.py#L232

jistr:
I kept getting blocked by the failures so i fixed the idempotence in PR #149. But update is still broken, it just doesn't get exercised right now.

os_migrate/plugins/module_utils/subnet.py
```diff
@@ -71,6 +71,13 @@ def _find_sdk_res(conn, name_or_id):
 
     @staticmethod
     def _update_sdk_res(conn, name_or_id, sdk_params):
+        # intermittently the subnet update fails when trying to update with a
+        # message that states the ip_version or network_id attributes cannot
+        # be updated.  See #140 for more information.
+        blacklist = ['network_id', 'project_id']
```
jistr, line 77: Is it worth doing this in Resource class? Then i wouldn't then call it `blacklist`, but perhaps `readonly_sdk_params`. And we could call a method which will filter those params out from `sdk_params` right before calling _update_sdk_res here:

https://github.com/os-migrate/os-migrate/blob/master/os_migrate/plugins/module_utils/resource.py#L175

WDYT?

jistr: We can also merge this fix first and then do the Resource subsequently if you want. And track it in a separate issue/card?

rbrady: I don't have a problem promoting this change into the base class.  I'll push an update to this patch in a few minutes.

os_migrate/plugins/module_utils/resource.py
```diff
@@ -228,6 +233,18 @@ def _needs_update(self, target):
...
+    def _remove_readonly_params(self, sdk_params):
+        """Remove readonly parameters from the `sdk_params` collection that
+        cannot be used for update.
+
+        Returns: filtered dict of parameters
+        """
+        for name in self.readonly_sdk_params:
+            if name in sdk_params:
+                sdk_params.pop(name)
+        return sdk_params
```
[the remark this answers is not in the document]
jistr: Since i know you're a Rust fan -- if we were writing this in Rust, it would outright force us to do (1) or even stricter variant of (2) with full deep copy and not reusing any internal data. What you've implemented here, or the "weak" copy variant i suggested as (2), would be refused by the compiler.

That makes me second guess my preference for (2) :smile: but i'm still fine with either (1) or (2).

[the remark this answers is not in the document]
jistr: Or full deep copy is fine too of course.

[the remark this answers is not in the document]
rbrady: I think option 1 is more simple, easier to read.  not sure if option two is worth the effort at this time.

PR #147, opened 2020-03-27
OOP sec group rules

ccamacho:
This is the third patch for the OOP refactor of the security groups and rules.

### DIFF
--- a/os_migrate/plugins/module_utils/security_group_rule.py
+++ b/os_migrate/plugins/module_utils/security_group_rule.py
@@ -4,79 +4,15 @@ __metaclass__ = type
 import openstack
 
 from ansible_collections.os_migrate.os_migrate.plugins.module_utils \
-    import const, exc, reference
-from ansible_collections.os_migrate.os_migrate.plugins.module_utils.serialization \
-    import set_sdk_params_same_name, set_ser_params_same_name
+    import const, reference, resource
 
 
-def security_group_rule_refs_from_sdk(conn, sdk_rule):
-    """Create a dict of name/id mappings for resources referenced from
-    OpenStack SDK Security Group Rule `sdk_rule`. Fetch any necessary information
-    from OpenStack SDK connection `conn`.
+class SecurityGroupRule(resource.Resource):
 
-    Returns: dict with names and IDs of resources referenced from
-    `sdk_rule` (only those important for OS-Migrate)
-    """
-    expected_type = openstack.network.v2.security_group_rule.SecurityGroupRule
-    if type(sdk_rule) != expected_type:
-        raise exc.UnexpectedResourceType(expected_type, type(sdk_rule))
-    refs = {}
+    resource_type = const.RES_TYPE_SECURITYGROUPRULE
+    sdk_class = openstack.network.v2.security_group_rule.SecurityGroupRule
 
-    # when creating refs from SDK Security Group Rule object, we copy IDs and
-    # query the cloud for names
-    refs['security_group_name'] = reference.security_group_name(
-        conn, sdk_rule['security_group_id'])
-    refs['remote_group_name'] = reference.security_group_name(
-        conn, sdk_rule['remote_group_id'])
-
-    return refs
-
-
-def security_group_rule_refs_from_ser(conn, ser_sec):
-    """Create a dict of name/id mappings for resources referenced from
-    OS-Migrage security group rule serialization `sdk_sec`. Fetch any necessary
-    information from OpenStack SDK connection `conn`.
-
-    Returns: dict with names and IDs of resources referenced from
-    `sdk_sec` (only those important for OS-Migrate)
-    """
-    if ser_sec[const.RES_TYPE] != const.RES_TYPE_SECURITYGROUPRULE:
-        raise exc.UnexpectedResourceType(
-            const.RES_TYPE_SECURITYGROUPRULE, ser_sec[const.RES_TYPE])
-    ser_params = ser_sec[const.RES_PARAMS]
-    refs = {}
-
-    # when creating refs from serialized security group, we copy names and
-    # query the cloud for IDs
-    refs['security_group_name'] = ser_params['security_group_name']
-    refs['security_group_id'] = reference.security_group_id(
-        conn, ser_params['security_group_name'])
-
-    return refs
-
-
-def serialize_security_group_rule(sdk_sec_rule, sec_refs):
-    """Serialize OpenStack SDK security group rule
-    `sdk_sec_rule` into OS-Migrate format.
-
-    Returns: Dict - OS-Migrate structure for Security group rule.
-    """
-    expected_type = openstack.network.v2.security_group_rule.SecurityGroupRule
-    if type(sdk_sec_rule) != expected_type:
-        raise exc.UnexpectedResourceType(expected_type, type(sdk_sec_rule))
-
-    resource = {}
-    params = {}
-    info = {}
-
-    resource[const.RES_PARAMS] = params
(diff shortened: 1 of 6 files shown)


os_migrate/plugins/module_utils/resource.py
```diff
@@ -158,15 +158,22 @@ def create_or_update(self, conn):
...
+                self._create_sdk_res(conn, sdk_params)
                 return True
+            return False  # no change done
         else:
-            self._create_sdk_res(conn, sdk_params)
-            return True
-        return False  # no change done
+            return False  # no change done
+            # self._create_sdk_res(conn, sdk_params)
+            # return True
```
jistr: For a resource without `name`, this method will never do antything. That's not correct. I would say, rather than editing the Resource implementation here, just fully override the `create_or_update` method in the SecurityGroupRule class. 
If you want insipration you can look at RouterInterface:

https://github.com/os-migrate/os-migrate/blob/684ff405ce7b8e44c51beb24d0b0baef154e56f7/os_migrate/plugins/module_utils/router_interface.py#L54

That resource also doesn't use a `name` as a key to look up existing one, rather it uses a combination of network_name + subnet_name + IP address:

https://github.com/os-migrate/os-migrate/blob/684ff405ce7b8e44c51beb24d0b0baef154e56f7/os_migrate/plugins/module_utils/router_interface.py#L87

RouterInterface is very similar to SecurityGroupRule, it also gets exported in bulk per one router, but it gets imported per-interface, like the export_security_group_rules and import_security_group_rule. Just the creation logic is a bit more complex in RouterInterface because it's done in 2 steps (create port, attach to router). However, structure-wise it might serve as a good pattern perhaps.

ccamacho: done :) I forgot to add my idenpotency check. And this was the place to add it.

os_migrate/plugins/module_utils/security_group_rule.py
```diff
@@ -4,132 +4,81 @@
...
-    """
-    res_type = ser_sec_rule.get(const.RES_TYPE, None)
-    if res_type != const.RES_TYPE_SECURITYGROUPRULE:
-        raise exc.UnexpectedResourceType(const.RES_TYPE_SECURITYGROUPRULE, res_type)
-
-    ser_params = ser_sec_rule[const.RES_PARAMS]
-    sdk_params = {}
+    params_from_refs = [
+        'security_group_name',
+        # FIXME: remote_group_name
```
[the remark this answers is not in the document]
ccamacho: done

os_migrate/plugins/module_utils/security_group_rule.py
```diff
@@ -4,132 +4,92 @@
...
+        sdk_params = self._to_sdk_params(refs)
+        result = {}
+        result['changed'] = False
+        try:
+            conn.network.create_security_group_rule(**sdk_params)
+            result['changed'] = True
+        except openstack.exceptions.ConflictException:
+            result['msg'] = "This security group rule already exists"
+
+        return result
```
jistr, line 58: This should return a boo whether the resource changed, not a result dict.

I see why you did this but IMO we should not rewrite the contract of the parent class. We should probably drop the 'msg' info here and just return False.

os_migrate/plugins/modules/import_security_group_rule.py
```diff
@@ -74,17 +74,8 @@ def run_module():
...
-
-    # The create security group rule method will return an exeption
-    # If it's already created.
-    try:
-        conn.network.create_security_group_rule(**secrule_params)
-        result['changed'] = True
-    except sdk.exceptions.ConflictException:
-        result['msg'] = "This security group rule already exists"
+    ser_secrule = security_group_rule.SecurityGroupRule.from_data(module.params['data'])
+    result = ser_secrule.create_or_update(conn)
```
jistr, line 78: And then here we'd have `result['changed'] =`

os_migrate/plugins/module_utils/security_group_rule.py
```diff
@@ -4,132 +4,92 @@
...
+    def _find_sdk_res(conn, name_or_id):
+        return conn.network.find_security_group_rule(name_or_id)
+
+    @staticmethod
+    def _refs_from_sdk(conn, sdk_res):
+        refs = {}
+        refs['security_group_id'] = sdk_res['security_group_id']
+        refs['security_group_name'] = reference.security_group_name(
+            conn, sdk_res['security_group_id'])
+        refs['remote_group_name'] = reference.remote_group_name(
```
jistr, line 74: Missing `remote_group_id`

os_migrate/plugins/module_utils/security_group_rule.py
```diff
@@ -4,132 +4,92 @@
...
-    sdk_params = {}
+    params_from_refs = [
+        'security_group_name',
+        'remote_group_name',
+    ]
 
-    set_sdk_params_same_name(sec_refs, sdk_params, [
-        'remote_group_id',
+    sdk_params_from_refs = [
         'security_group_id',
```
jistr, line 39: Missing `remote_group_id`

os_migrate/plugins/module_utils/security_group_rule.py
```diff
@@ -4,132 +4,92 @@
...
+            conn, sdk_res['security_group_id'])
+        refs['remote_group_name'] = reference.remote_group_name(
+            conn, sdk_res['remote_group_id'])
+        return refs
+
+    def _refs_from_ser(self, conn):
+        refs = {}
+        refs['security_group_name'] = self.params()['security_group_name']
+        refs['security_group_id'] = reference.security_group_id(
+            conn, self.params()['security_group_name'])
```
jistr, line 82: Missing `remote_group_name` and `remote_group_id`

jistr: Still missing

os_migrate/plugins/module_utils/security_group_rule.py
```diff
@@ -4,132 +4,92 @@
...
+        refs = self._refs_from_ser(conn)
+        sdk_params = self._to_sdk_params(refs)
+        try:
+            conn.network.create_security_group_rule(**sdk_params)
+            return True
+        except openstack.exceptions.ConflictException:
+            # TODO: Log that the security group rule already exists
+            return False
+
+    @staticmethod
```
jistr, line 58: Nit: I don't think this method is used anywhere, can be deleted. I didn't realize this in the previous pass sorry.

os_migrate/plugins/module_utils/security_group_rule.py
```diff
@@ -4,132 +4,92 @@
...
+    def _create_sdk_res(conn, sdk_params):
+        return conn.network.create_security_group_rule(**sdk_params)
+
+    @staticmethod
```
jistr, line 62: Ditto re deleting.

os_migrate/plugins/module_utils/security_group_rule.py
```diff
@@ -4,132 +4,92 @@
...
+        return refs
+
+    def _refs_from_ser(self, conn):
+        refs = {}
+        refs['security_group_name'] = self.params()['security_group_name']
+        refs['security_group_id'] = reference.security_group_id(
+            conn, self.params()['security_group_name'])
+        return refs
+
+    @staticmethod
```
jistr, line 85: Ditto re deleting, also this returns a dict which breaches the parent contract.

jistr: _update_sdk_res still can be deleted

PR #148, opened 2020-03-30
Convert the rest of modules to new auth

### LINKED ISSUE #143: Implement new authentication mechanism

jistr:
Closes #143

### DIFF
--- a/.github/workflows/consistency-functional.yml
+++ b/.github/workflows/consistency-functional.yml
@@ -193,8 +193,7 @@ jobs:
               --src devstack \
               --dst devstack-alt \
               > tests/func/auth.yml && \
-          sed -i -e "s/ devstack:/ testsrc:/" tests/func/clouds.yaml && \
-          sed -i -e "s/ devstack-alt:/ testdst:/" tests/func/clouds.yaml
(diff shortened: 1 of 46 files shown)


jistr:
@ccamacho @rbrady Can we merge this (and dependencies #149 and #144) relatively fast please unless there are some major issues with the patches? This patch chain edits pretty much the whole project in breadth (46 files currently) so there is a large potential for collisions if we leave this on review for extended amount of time.

Once we merge this PR, we are completely switched to the new auth mechanism which should be suitable for workload migration, and the old mechanism is cleaned up to ensure we don't start depending on it somewhere by accident.

jistr:
@ccamacho @rbrady This is probably one of the patches where at least an ack from both of you (at least skim-through if you don't have much time) would be good before merging, since it's a significant change of the auth pattern in the whole project.

jistr:
PR #144 is actually where the new pattern is implemented, so that one probably needs a double-ack. This PR just finishes the work across the whole project.

Issue #156, opened 2020-04-03
Tests test wrong code when there's higher version collection archive in releases dir


jistr:
When we run tests, we install the collection. The logic looks for `releases/os_migrate-os_migrate-latest.tar.gz`. This is a symlink which is created by pointing to whatever is the highest version tar.gz archive found in the releases dir.

Recently we went "back in versions" before releasing first code to Galaxy. So locally we started building v0.0.1 but in the releases directory we had stale v0.1.0. Tests never tested our code changes, they always installed the stale release.

While the above is a scenario which is unlikely to repeat, we may be e.g. fixing a bug on stable 1.0 branch in the future, while already having worked on 2.0 release. In this scenario the tests would locally run always the 2.0 code even when working on 1.0 branch. This is a realistic bug scenario.

We should probably stop creating -latest.tar.gz and start creating -last-build.tar.gz and always update it based on what was built, and not based on what is highest version.

ccamacho:
@jistr when we build the collection for the tests there should be only one "active" collection, we should be testing always the lates which is also the only one.

PR #157, opened 2020-04-03
Do not ls, use always current version as latest

ccamacho:
This might hit us in our dev envs, we should push a fix for this asap.

Closes: #156


ccamacho:
To avoid issues, we can keep the symlink as is used in the makefile

PR #159, opened 2020-04-08
Add a logging interface

ccamacho:


os_migrate/plugins/module_utils/logger.py
```diff
@@ -0,0 +1,34 @@
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+import logging
+from logging.handlers import TimedRotatingFileHandler
+FORMATTER = logging.Formatter("%(asctime)s - %(name)s - "
+                              "%(levelname)s - %(message)s")
+LOG_FILE = "/tmp/os_migrate.log"
```
ccamacho, line 8: /var/tmp vs /tmp ?? I think storing in /var/log will be a bit more problematic as we will need root permissions to write there.

ccamacho: Do we need the logs after a reboot? I /tmp is a ramdisk it will go away but maybe we wont need them anymore?

rbrady: If you're going to use tmp, wouldn't it be better to use gettempdir() [1] so that we use the users preferred temporary directory and not assume it's /tmp ?

[1] https://docs.python.org/3.7/library/tempfile.html#tempfile.gettempdir

os_migrate/plugins/module_utils/security_group.py
```diff
@@ -26,6 +26,7 @@ class SecurityGroup(resource.Resource):
 
     @classmethod
     def from_sdk(cls, conn, sdk_resource):
+        logger.get_logger("from_sdk").debug("Fetching data from SDK")
```
jistr, line 29: We probably need to discuss this more, but i don't think our primary focus should be on this kind of logging? IMO primarily we should be saving the Ansible run logs, so probably using something like: https://docs.ansible.com/ansible/latest/reference_appendices/config.html#default-log-path ?

BTW this is M4 task in Jira board. We can discuss if we think the priorities should change on anything, but i'm not convinced we should be focusing on this until all M3s are either done or have assignees working on them?

os_migrate/plugins/module_utils/logger.py
jistr: Yea we should preserve logs through reboots i think. Ideally we'd get our /var/log location but it might require having an RPM, so perhaps /var/tmp with proper ownership like Ryan is suggesting is better. If we use mkdtemp we should proably be able to choose between /tmp and /var/tmp explicitly, and we should also include user name in prefix to make sure different users on the machine don't try to write into the same dir (which would fail with proper permissions).

PR #160, opened 2020-04-08
Adds Router Idempotence Fuctional Test

rbrady:
This patch adds a test to ensure idempotence for the router resource.

### DIFF
--- a/tests/func/idempotence/all.yml
+++ b/tests/func/idempotence/all.yml
@@ -21,3 +21,11 @@
       tags:
         - test_security_group
   tags: always
+
+- name: Include router tasks
+  include_tasks: router.yml
+  args:
+    apply:
+      tags:
+        - test_router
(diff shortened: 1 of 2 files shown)


rbrady:
Pushed the current state as progress.  Import is failing right now.  Just saw #166 so I'll see if that fixes the issue I'm seeing.

jistr:
@rbrady Yea there are still issues with Router, external gateway not getting added, which would probably break the idempotence test too. My first fix attempt didn't work, i'll try another way in a PR separate from #166. (I also commented at https://github.com/os-migrate/os-migrate/pull/166#issuecomment-613935297 ).

jistr:
This should get us closer, there may still be some serialization ordering issues as with other resources but i'm hoping there are no functional bugs left: https://github.com/os-migrate/os-migrate/pull/168

PR #162, opened 2020-04-08
Basic workload information export.

mrnold:
Export instance information from a source OpenStack cloud to
workloads.yml. Actual workload migration will require saving more
properties than this, but they will get filled in later.


os_migrate/plugins/module_utils/server.py
```diff
@@ -0,0 +1,24 @@
...
+
+    resource_type = const.RES_TYPE_SERVER
+    sdk_class = openstack.compute.v2.server.Server
+
+    info_from_sdk = [
+        'addresses',
+        'id',
+        'name',
+        'status',
+    ]
```
jistr, line 20: At least `name` should be in `params_from_sdk`, perhaps `addresses` should be there too. The difference between `info` and `params` is:

* info is something that we don't want or cannot "make the same" in the destination cloud. So `id` should remain in info (typically any UUIDs can't be the same), and `status` probably too unless/until we want to support creation of instances in the destination cloud without starting them up.

* params are things that we do want transferred/copied to the destination cloud. These typically also get looked at when we're making sure the import behavior is idempotent. I.e. when we re-run import playbooks, things that already got imported before are skipped. For workloads, the idempotence would probably only be looking at VM names - "i see a VM with this name is already in the destination cloud, skip the import of this one".

The workload migration will probably not "fit existing molds" in some ways so we will probably be ironing things out over time and finding new patterns, but at the very least `name` should move into `params` (so `params_from_sdk`). Other than this, the PR is ready to merge i think. Thanks!

os_migrate/tests/unit/test_workload.py
```diff
@@ -0,0 +1,68 @@
+from __future__ import (absolute_import, division, print_function)
```
jistr, line 1: I wanted to comment we should have a functional export test but then realized GH Actions won't let us create an instance. So we'll only test this in end-to-end tests eventually (cc FYI @ccamacho ).

os_migrate/plugins/module_utils/server.py
mrnold: Thanks for the clarification. I am not sure if we can preserve addresses exactly on the destination cloud, but I can look into it if that's the way people are going to expect this to work.

jistr:
@mrnold can you try rebasing on the latest master please? We probably need that to make the CI pass.

PR #166, opened 2020-04-14
Router fixes

jistr:


jistr:
I think there will still be an issue with external gateway, it doesn't seem to get added onto the imported Router. I'm trying to fix that but haven't been successful just yet. I fixed _refs_from_ser for Router, but it still doesn't add the gateway. I may have to add it in a separate request maybe... That would probably make sesnse as a separate PR, so i'm un-WIPing this one, ready to merge.

PR #171, opened 2020-04-16
Save ether_type for security group rules

jistr:
This is a fundamental property of the rule, specifies whether the rule
applies to IPv4 or IPv6 traffic.


jistr:
@rbrady I wonder if this helps with the idempotence patch? I noticed 2 serialized rules in the tests which looked the same, thought that's weird so started digging and discovered one is for IPv4 and the other for IPv6, but it wasn't in the serialized info.

rbrady:
I rebased the patch on this and the error with idempotence is still failing on verifying the export file did not change.

Issue #21, opened 2020-01-16
[M3 gate issue] all imports/exports idempotent?


jistr:
This is so that M1 milestone only gets closed after we verify that export/import of all resources is idempotent.

rbrady:
All resources are currently idempotent.  Not sure of the workload migration idempotence at this time.

jistr:
Thanks, workload migration is its own beast, i'd say let's tackle it when we have the e2e CI for it.

jistr:
Reported https://github.com/os-migrate/os-migrate/issues/185 and we can close this one.

PR #176, opened 2020-04-23
Adds test for Router Interface idempotence

rbrady:
This patch adds a test to ensure router interface idempotence.

### DIFF
--- a/os_migrate/plugins/module_utils/router_interface.py
+++ b/os_migrate/plugins/module_utils/router_interface.py
@@ -81,7 +81,6 @@ class RouterInterface(resource.Resource):
             conn.network.add_interface_to_router(
                 refs['device_id'], port_id=port['id'])
             changed = True
-
         return changed
 
     def _find_port_by_subnet_ips(self, conn, refs):
@@ -151,7 +150,7 @@ class RouterInterface(resource.Resource):
         for fixed_ip in params['fixed_ips_names']:
             refs['fixed_ips'].append({
                 'ip_address': fixed_ip['ip_address'],
-                'subnet_name': reference.subnet_id(conn, fixed_ip['subnet_name']),
+                'subnet_id': reference.subnet_id(conn, fixed_ip['subnet_name']),
             })
         refs['device_name'] = params['device_name']
         refs['device_id'] = reference.router_id(conn, params['device_name'])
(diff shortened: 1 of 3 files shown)


rbrady:
This latest push is just the progress so far.  It's currently throwing openstack.exceptions.ConflictException when running through the import in the idempotence test.  I'll address this exception in a change tomorrow.

rbrady:
The create_or_update method has been modified to handle the exception, but the test isn't likely checking the right data to ensure idempotency (it's looking at routers.updated_at)

os_migrate/plugins/module_utils/router_interface.py
```diff
@@ -58,30 +58,32 @@ def create_or_update(self, conn):
...
-                refs['device_id'], port_id=port['id'])
-            changed = True
 
+            # 2. ensure the port is attached to the router by `router_name`
+            if port['device_owner'] != 'network:router_interface' \
+               and port['device_id'] != refs['device_id']:
+                conn.network.add_interface_to_router(
+                    refs['device_id'], port_id=port['id'])
+                changed = True
+        except openstack.exceptions.ConflictException:
```
jistr, line 84: There shouldn't be any conflict exception here:

* we're looking up the port by IPs, if it exists, we don't create it

* we're looking if it's assigned to the router already, if yes, we don't try to assign it again

The method should be a no-op, something's not working. I think we should figure out what's being executed that shouldn't be executed, rather than catching the exception.

rbrady: The error message output is here: https://gist.github.com/rbrady/ff2b8f41cb3a6738599102478d900078

rbrady:
closes #21

os_migrate/plugins/module_utils/router_interface.py
```diff
@@ -151,7 +150,7 @@ def _refs_from_ser(self, conn):
         for fixed_ip in params['fixed_ips_names']:
             refs['fixed_ips'].append({
                 'ip_address': fixed_ip['ip_address'],
-                'subnet_name': reference.subnet_id(conn, fixed_ip['subnet_name']),
+                'subnet_id': reference.subnet_id(conn, fixed_ip['subnet_name']),
```
jistr, line 153: Damn :). Thanks for the fix.

PR #180, opened 2020-05-05
Allow better workload debug logging and progress tracking

mrnold:
Split the workload migration tasks into two stages, so that the first stage can display log locations before going silent for the bulk data transfer. It ends up looking like this:

```
TASK [import_workloads : preliminary setup for workload import] ************************************************************
changed: [localhost]

TASK [import_workloads : debug] ********************************************************************************************
ok: [localhost] => {
    "msg": [
        "second-migration-vm local stdin/stderr: /home/migrate/lab-migrate-data/second-migration-vm.log",
        "second-migration-vm log file: cloud-user@10.19.2.137:/tmp/v2v-6aHvCz/log/uci/virt-v2v-wrapper.log",
        "second-migration-vm progress state file: cloud-user@10.19.2.137:/tmp/v2v-6aHvCz/lib/uci/state.json"
    ]
}

TASK [import_workloads : import one workload] ******************************************************************************
changed: [localhost]
```

### DIFF
--- a/os_migrate/plugins/module_utils/workload_common.py
+++ b/os_migrate/plugins/module_utils/workload_common.py
@@ -1,0 +1,23 @@
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+import subprocess
+
+
+def ssh_preamble(address, key_path):
+    """ Common options to SSH into a conversion host. """
+    return [
+        'ssh',
+        '-i', key_path,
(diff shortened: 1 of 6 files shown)


mrnold:
I should also mention up-front that this creates files named [VM name].log directly in the data directory. I'm not sure if it would be preferable to put these in a separate directory, or a sub-directory, or have another top-level user variable.

PR #181, opened 2020-05-11
Include the workload migration into e2e jobs

ccamacho:


ccamacho:
It's passing, but it's a lie xD almost there

os_migrate/plugins/modules/os_conversion_host_info.py
```diff
@@ -116,7 +116,13 @@ def main():
...
         if not server:
             module.fail_json(msg='Conversion host ' + srv + ' not found!')
-        conversion_host['address'] = server.accessIPv4
+        if server.public_v4 == "":
+            module.fail_json(msg='Conversion host address cant be empty \n' + str(server))
+        # The value of the server.accessIPv4 property is not defined by OpenStack,
+        # and may not be consistent across vendors.
+        # TODO:DISCUSSION Should we use server.public_v4 instead?
+        # conversion_host['address'] = server.accessIPv4
+        conversion_host['address'] = server.public_v4
```
jistr, line 125: I'm not familiar with these properties but just want to comment here to mark it as "this warrants a closer look before merging". Specifically i wonder if the "public_v4" name is generic, or if it is just generated because our external network happens to be named "public", in which case this would break the migration for other environments.

os_migrate/roles/import_workloads/tasks/workload.yml
```diff
@@ -47,5 +47,9 @@
     when: prelim.changed
 
   rescue:
-    - debug:
+    - name: give some time to debug this!
```
jistr, line 50: Should the wait_for go into the e2e test instead -- let the ansible playbook fail as normal, and if it fails, we keep the env around for some time?

I think we should keep the change from 'debug' to 'fail' here though. The rescue+debug pattern could result in unnoticed errors?

Makefile
```diff
@@ -73,8 +73,20 @@ test-e2e: reinstall
...
 		-e os_migrate_src_validate_certs=False \
 		-e os_migrate_dst_validate_certs=False \
+		-e os_migrate_src_conversion_host_name=osm_uch_src \
+		-e os_migrate_dst_conversion_host_name=osm_uch_dst \
+		-e os_migrate_src_conversion_host_flavor=m1.medium \
+		-e os_migrate_dst_conversion_host_flavor=m1.medium \
+		-e os_migrate_src_client_key=~/.ssh/id_rsa.pub \
+		-e os_migrate_dst_client_key=~/.ssh/id_rsa.pub \
+		-e os_migrate_conversion_host_key=~/.ssh/id_rsa \
+		-e os_migrate_conversion_host_image=v2v-conversion-host-appliance-devel.qc2 \
```
jistr, line 89: Eventually we may want to move to passing a file here so that these params are generated by the CI environment and not hardcoded in the Makefile. Let's do it in a subsequent patch though, no need to churn on this one.

os_migrate/plugins/modules/import_workload_prelim.py
```diff
@@ -267,14 +267,47 @@ def run_module():
...
+                             'Error was: ' + e, **result)
+
     # Do not convert source conversion host!
     if info['id'] == module.params['src_conversion_host']['id']:
         module.exit_json(skipped=True, skip_reason='Skipping conversion host.',
                          **result)
 
     # Assume an existing VM with the same name means it was already migrated.
     # Not necessarily true, but force the operator to delete it if needed.
-    if len(list(conn.compute.servers(server_name, **module.params['dst_filters']))) > 0:
```
jistr, line 277: Ok let's go back to the previous query and once the CI changes are merged, we'll investigate what went wrong here.

jistr:
Yay for e2e CI! Nice work!

PR #182, opened 2020-05-13
Implement export filter for resources in current project (for networks)

jistr:
This adds the reusable code for fetching current project ID and
setting a general export filter with that project ID by
default. Networks export is amended to make use of this filtering
parameter.


jistr:
The e2e job failed with:

 msg: 'An unhandled exception occurred while running the lookup plugin ''file''. Error was a <class ''ansible.errors.AnsibleError''>, original message: could not locate file in lookup: /root/os_migrate/tests/func/tmpdata/networks.yml'

That looks unrelated...

This PR should be ready to merge.

PR #187, opened 2020-05-19
Project filters for workload import

jistr:


jistr:
Just the last commit is relevant, i can rebase after #184 gets merged to slim down the diff.

PR #190, opened 2020-05-26
Adds additional user docs

rbrady:
This patch adds more data to the repo README and updates the
the user doc to remove clouds.yml info and replace it with
the newer auth vars.

* Note this patch is still WIP as I'm trying to figure out the
best way to adapt the blog post to the user docs as well.

### DIFF
--- a/doc/user/README.md
+++ b/doc/user/README.md
@@ -30,53 +30,8 @@ It is recommended to install releases from Ansible Galaxy.
 ansible-galaxy collection install os_migrate.os_migrate
 ```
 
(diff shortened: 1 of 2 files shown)


jistr:
@rbrady should we merge? After we merge this, we may want to split the user docs into 3 parts:

* Getting started  (this is what we currently have)

* Migration walkthrough (what you suggested on the call recently, full end-to-end example, if i understood correctly)

* Variables reference (what i suggested on the call recently, a reference of what variables are available for setting and what they do)

rbrady:
I was adding the end to end walk through to this patch, but happy to merge this and add a follow up patch.

PR #191, opened 2020-05-27
Reintroduce project filtering into workload migration idempotence check

jistr:


ccamacho:
Here you have the filter skipping the migration
https://gitlab.cee.redhat.com/os-migrate/bot/-/jobs/1649205#L5359

jistr:
@ccamacho yea i've seen. I've added some debug info to decipher what's going on.

PR #200, opened 2020-06-04
Workload migration without virt-v2v-wrapper

mrnold:
Move the workload migration code out of virt-v2v-wrapper and into a few new ansible stages. This allows the use of standard cloud images as conversion hosts, with a few extra dependencies: basically just qemu-img, nbdkit or qemu-nbd, and optionally virt-sparsify. All OpenStack API manipulations are performed on the migration host, and everything that needs to be done on the conversion hosts is run over SSH.


os_migrate/plugins/modules/import_workload_prelim.py
```diff
@@ -319,7 +219,9 @@ def run_module():
     # Assume an existing VM with the same name means it was already migrated.
     # Not necessarily true, but force the operator to delete it if needed.
 
-    if len(list(conn.compute.servers(details=False, name=server_name, **module.params['dst_filters']))) > 0:
+    # TODO:FIXME: The following filter returns > 0 and there is no VM
+    # if len(list(conn.compute.servers(details=False, name=server_name, **module.params['dst_filters']))) > 0:
+    if conn.search_servers(server_name):
```
jistr, line 224: The PR is probably just not based on the latest master branch, we shouldn't be reverting this bit i think?

os_migrate/roles/import_workloads/tasks/workload.yml
```diff
@@ -8,42 +8,91 @@
...
+      auth_type: "{{ os_migrate_src_auth_type|default(omit) }}"
+      region_name: "{{ os_migrate_src_region_name|default(omit) }}"
+      validate_certs: "{{ os_migrate_src_validate_certs|default(omit) }}"
+      ca_cert: "{{ os_migrate_src_ca_cert|default(omit) }}"
+      client_cert: "{{ os_migrate_src_client_cert|default(omit) }}"
+      client_key: "{{ os_migrate_src_client_key|default(omit) }}"
+      conversion_host:
+        "{{ os_src_conversion_host_info.openstack_conversion_host }}"
+      data: "{{ item }}"
+      log_file: "{{ os_migrate_data_dir }}/{{ prelim.server_name }}.log"
```
jistr, line 36: Later i think we'll want to separate these from the yml definitions of resources (perhaps into a subdir), but this may be best done later, no need to disturb this PR.

mrnold:
> Thanks, this is some clean code! I didn't dive into all the details to form a full understanding of every corner, but high level this looks sensible and nicely structured. I'm curious if it would "just work" with the heavyweight conversion host that we're presently deploying in CI. Triggering an e2e job.

It won't work with the current conversion host, because all the requirements are wrapped up inside the v2v-conversion-host container. It does work with the older conversion host appliance from OSP14 though.

os_migrate/plugins/modules/import_workload_prelim.py
mrnold: Right, I will take this out.

jistr:
So i'm wondering how to best merge this so that CI is working. Maybe it's time to add playbooks for deployment of conversion hosts and perhaps building of conversion host image, and add them both into the CI?   @ccamacho @rbrady

jistr:
I'll try to tackle the deployment first, and maybe we can start with CentOS which can probably install deps easily? @mrnold what would we need installed in the conversion host? Just python3-openstacksdk?

mrnold:
The base dependency list is: nbdkit nbdkit-basic-plugins qemu-img libguestfs-tools

Those will pull in a bunch of other packages. This pull request moves all the active code into os_migrate, and all the work that needs to be done on the conversion hosts is done through SSH. That means there is no dependency on openstacksdk, although it does produce some ugly command line contortions.

jistr:
Thanks :)

mrnold:
The other part of deployment is that the destination conversion host must have an SSH key that is authorized to access the source conversion host without a passphrase. With the UCI, this works because virt-v2v-wrapper sets up an SSH agent with the given key, so the whole conversion process runs under that agent. So far I have just been creating a new key pair and copying the public key to the source conversion host, but there's probably much nicer ways of doing this.

PR #210, opened 2020-06-15
Adds Functional Test for Network Update

rbrady:
This patch adds the code to test the update of the network resource
and will ultimately include the fix for the failing update once
I find the root cause. :)

Note this patch is WIP.

### DIFF
--- a/tests/func/update/network.yml
+++ b/tests/func/update/network.yml
@@ -1,0 +1,58 @@
+- name: change item in file for update test
+  lineinfile:
+    dest: "{{ os_migrate_data_dir }}/networks.yml"
+    regexp: "description: ''"
+    line: "    description: 'osm_net_updated'"
+    state: present
+    backup: yes
+    backrefs: yes
+
+- name: scan available networks after before test
+  os_networks_info:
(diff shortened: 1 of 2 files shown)


tests/func/update/network.yml
```diff
@@ -0,0 +1,47 @@
...
+    client_cert: "{{ os_migrate_dst_client_cert|default(omit) }}"
+    client_key: "{{ os_migrate_dst_client_key|default(omit) }}"
+  register: dst_networks_info_after
+
+- name: ensure description for osm_net changed
+  assert:
+    that:
+      - dst_networks_info_before['openstack_networks'][0].description != None
+      - "dst_networks_info_before['openstack_networks'][0]['description'] \
+         != dst_networks_info_after['openstack_networks'][0]['description']"
```
jistr, line 42: We might even check for the actual value `osm_net_updated` here but not a blocker.

jistr:
Ready to land? Might need a rebase though. The func test patches conflict with each other so unless we make them one chain, we'll need to rebase after each merge.

rbrady:
Changing the asserts to be more specific and chaining the patches.  This patch is #1 in the chain atm.

PR #212, opened 2020-06-16
Adjust playbooks variables

ccamacho:


tests/scenario_variables.yml
```diff
@@ -0,0 +1,18 @@
+# This is the default scenario for running the end-to-end tests.
```
jistr, line 1: Since this file is specific to e2e test, we should probably move it to the e2e folder.

ccamacho: agreed

PR #213, opened 2020-06-16
Adds Update Functional Tests for Subnets

rbrady:
This patch adds the code to test the update functionality of the
subnet resources.

Note: this patch is WIP

### DIFF
--- a/tests/func/update/subnet.yml
+++ b/tests/func/update/subnet.yml
@@ -1,0 +1,58 @@
+- name: change item in file for update test
+  lineinfile:
+    dest: "{{ os_migrate_data_dir }}/subnets.yml"
+    regexp: "    description: ''"
+    line: "    description: 'osm_subnet_updated'"
+    state: present
+    backup: yes
+    backrefs: yes
+
+- name: scan available subnets after before test
+  os_subnets_info:
+    filters: "{{ os_migrate_dst_filters }}"
+    auth: "{{ os_migrate_dst_auth }}"
+    auth_type: "{{ os_migrate_dst_auth_type|default(omit) }}"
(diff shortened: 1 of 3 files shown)


rbrady:
This code should not fail for ip_version or cidr being read-only.  Neutron API lists project_id and network_id as unable to update:

https://github.com/openstack/neutron/blob/master/neutron/objects/subnet.py#L232

jistr:
I think we have a bug here:

https://github.com/os-migrate/os-migrate/blob/5e9b3aad8d9d73dbdd4c760715a0ebfe791970cd/os_migrate/plugins/module_utils/resource.py#L198
https://github.com/os-migrate/os-migrate/blob/5e9b3aad8d9d73dbdd4c760715a0ebfe791970cd/os_migrate/plugins/module_utils/subnet.py#L75

The 2nd parameter for _update_sdk_res should be an actual subnet instance, or its ID, not its name. We should fix the _update_sdk_res methods parameter naming throughout the project.

tests/func/update/subnet.yml
```diff
@@ -0,0 +1,47 @@
...
+    client_cert: "{{ os_migrate_dst_client_cert|default(omit) }}"
+    client_key: "{{ os_migrate_dst_client_key|default(omit) }}"
+  register: dst_subnets_info_after
+
+- name: ensure description for osm_subnet changed
+  assert:
+    that:
+      - dst_subnets_info_before['openstack_subnets'][0].description != None
+      - "dst_subnets_info_before['openstack_subnets'][0]['description'] \
+         != dst_subnets_info_after['openstack_subnets'][0]['description']"
```
jistr, line 42: We might check for the actual value `osm_subnet_updated` but not a blocker.

PR #215, opened 2020-06-18
Router update test

rbrady:


tests/func/update/router.yml
```diff
@@ -0,0 +1,47 @@
...
+    client_cert: "{{ os_migrate_dst_client_cert|default(omit) }}"
+    client_key: "{{ os_migrate_dst_client_key|default(omit) }}"
+  register: dst_routers_info_after
+
+- name: ensure description for osm_router changed
+  assert:
+    that:
+      - dst_routers_info_before['openstack_routers'][0].description != None
+      - "dst_routers_info_before['openstack_routers'][0]['description'] \
+         != dst_routers_info_after['openstack_routers'][0]['description']"
```
jistr, line 42: We might check for the actual value `osm_router_updated` but not a blocker.

PR #216, opened 2020-06-22
Adds Functional Update Test for Security Groups

rbrady:
This patch adds a test to ensure the updates for security group
resources is working properly.

NOTE: WIP for a template failure in the assert.

### DIFF
--- a/tests/func/update/security_group.yml
+++ b/tests/func/update/security_group.yml
@@ -1,0 +1,56 @@
+- name: change item in file for update test
+  lineinfile:
+    dest: "{{ os_migrate_data_dir }}/security_groups.yml"
+    regexp: "description: OSM security group"
+    line: "    description: 'OSM security group updated'"
+    state: present
+    backup: yes
+    backrefs: yes
+
+- name: scan available security_groups after before test
+  os_migrate.os_migrate.os_security_groups_info:
+    filters: "{{ os_migrate_dst_filters }}"
+    auth: "{{ os_migrate_dst_auth }}"
+    auth_type: "{{ os_migrate_dst_auth_type|default(omit) }}"
+    region_name: "{{ os_migrate_dst_region_name|default(omit) }}"
+    validate_certs: "{{ os_migrate_dst_validate_certs|default(omit) }}"
+    ca_cert: "{{ os_migrate_dst_ca_cert|default(omit) }}"
+    client_cert: "{{ os_migrate_dst_client_cert|default(omit) }}"
+    client_key: "{{ os_migrate_dst_client_key|default(omit) }}"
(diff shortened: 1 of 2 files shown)


tests/func/update/security_group.yml
```diff
@@ -0,0 +1,53 @@
...
+
+- name: updated description for osm_security_group
+  debug:
+    msg: "{{ (dst_sg_info_after['openstack_security_groups']
+      | json_query(\"[?name=='osm_security_group'].description\"))[0] }}"
+
+- name: ensure description for osm_security_group changed
+  assert:
+    that:
+      - dst_sg_info_before['openstack_security_groups'] | json_query(\"[?name=='osm_security_group'].description\")[0] == 'OSM security group'
```
jistr, line 50: Maybe we can instead of:
```
json_query(\"[?name=='osm_security_group'].description\")[0]
```
this:
```
json_query("[?name=='osm_security_group'].description")[0]
```

The above rang a bell with me, i've hit a similar issue with escaping earlier. The issue is that `debug: msg` parameter expects a string so it needs "{{ }}" use a jinja expression. But `assert: that` already expects an expression, so we don't wrap it in double quotes so we don't need to escape any double quotes used inside.

rbrady:
dropping the "\" out from the assert just changed the error message output slightly.  Instead of complaining about the "\" char, it now says the "[" char just before that escape char is expected at the end of the sequence:

'The conditional check ''dst_sg_info_before[''openstack_security_groups''] | json_query("[?name==''osm_security_group''].description")[0] == ''OSM security group'''' failed. The error was: template error while templating string: expected token ''end of statement block'', got ''[''. String: {% if dst_sg_info_before[''openstack_security_groups''] | json_query("[?name==''osm_security_group''].description")[0] == ''OSM security group'' %} True {% else %} False {% endif %}'

jistr:
It looks like you also changed the single quotes around osm_security_group into double quotes, if the error message represents that well? That wasn't in my suggestion above. I just suggested the backslash removal (removal of escaping).

rbrady:
I ended up looking into the error messages and found a github bug report[1] where someone had worked around it by using set_fact, because of a variable expansion issue in jinja2.  I looked at the source for ansible in the template/__init__.py file to review the comments in the _escape_backslashes source about multiple passes for escaping.[2]

In any case, using the set_fact seemed to get around the jinja2 template error message I was receiving and now the test is able to filter for a specific resource instead of operating on a list of resources that lacks a deterministic order.

[1]https://github.com/ansible/ansible/issues/14172
[2]https://github.com/ansible/ansible/blob/devel/lib/ansible/template/__init__.py#L130-L147

PR #218, opened 2020-06-26
Prepare release 0.4.0

jistr:
Mainly utilizes a new workload migration approach with new conversion
host and conversion host deployment/deletion playbooks.


ccamacho:
All the strings are replaced correctly LGTM! Ship it 0.4.0!

PR #226, opened 2020-06-30
Adds Walkthrough Document

rbrady:
This patch adds documentation for executing a parallel cloud migration
using os-migrate.

Co-authored-by: Jiri Stransky <jistr@redhat.com>
Co-authored-by: Matthew Arnold <marnold@redhat.com>


rbrady:
There are some things still needing to be addressed:

- do we change everything RHOS specific to OpenStack?

- do we reflect any of the recent changes around UCH?

ccamacho:
Yeahp, as a first iteration is OK. Still, as it is an "upstream project" we should kind of try making it flavor agnostic. Also, we can speak about the possible APIs incompatibilities we are experiencing and we can adjust the new conversion host parameters in a follow up if you wish.

jistr:
We can merge it first and then make edits, but @rbrady i'd say "yes" to both your questions, we should make those edits IMO.

I was also wondering if we want this as a high level walkthrough, or a full practical walk through even with commands. The pre-workload migration part is more the former, while the workload migration is more of the latter. We should either add detail to the pre-workload migration part or remove detail from the workload migration part i think.

PR #238, opened 2020-07-14
Adds Role to Install Conversion Host Content

rbrady:
This patch adds a role to install content for the conversion host
conditionally.  The role currently supports CentOS and RHEL but
is written in a way that should be simple to add additional OS
support if needed in the future.

[ccamacho]: We will merge this and iterate over, there is one simple change to make it work.

### DIFF
--- a/os_migrate/roles/conversion_host_content/tasks/main.yml
+++ b/os_migrate/roles/conversion_host_content/tasks/main.yml
@@ -1,0 +1,18 @@
+- name: display host os
+  debug: msg="Conversion Host OS is {{ ansible_distribution }}."
+
+- name: install epel release for CentOS
+  yum:
+    name: epel-release
+    state: present
+  when: ansible_distribution in ['CentOS']
+
+- name: install content
+  yum:
+    name:
+      - nbdkit
+      - nbdkit-basic-plugins
+      - qemu-img
+      - libguestfs-tools
+    state: present
+  when: ansible_distribution in ['CentOS', 'RedHat']
--- a/os_migrate/playbooks/deploy_conversion_hosts.yml
+++ b/os_migrate/playbooks/deploy_conversion_hosts.yml
@@ -52,3 +52,17 @@
         name: os_migrate.os_migrate.conversion_host
         tasks_from: link_insert_private_key.yml
       when: os_migrate_link_conversion_hosts|default(true)|bool
+
+- hosts: os_migrate_conv_src
+  tasks:
+    - include_role:
+        name: os_migrate.os_migrate.conversion_host_content
+      when: os_migrate_conversion_host_content_install|default(true)|bool
+
+- hosts: os_migrate_conv_dst
+  become_user: root
+  become: yes
+  tasks:
+    - include_role:
+        name: os_migrate.os_migrate.conversion_host_content
+      when: os_migrate_conversion_host_content_install|default(true)|bool
--- a/os_migrate/roles/conversion_host/defaults/main.yml
+++ b/os_migrate/roles/conversion_host/defaults/main.yml
@@ -25,3 +25,5 @@ os_migrate_conversion_host_name: os_migrate_conv
 os_migrate_conversion_image_name: os_migrate_conv
 
 os_migrate_conversion_ssh_user: cloud-user
+
+os_migrate_conversion_host_content_install: true
--- a/os_migrate/roles/conversion_host_content/README.md
+++ b/os_migrate/roles/conversion_host_content/README.md
@@ -1,0 +1,2 @@
+Please refer to
+[OS Migrate user docs](https://github.com/os-migrate/os-migrate/blob/master/doc/user/README.md).
--- a/os_migrate/roles/conversion_host_content/tasks/centos.yml
+++ b/os_migrate/roles/conversion_host_content/tasks/centos.yml
@@ -1,0 +1,13 @@
+- name: install epel release
+  yum:
+    name: epel-release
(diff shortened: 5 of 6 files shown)


os_migrate/roles/conversion_host/tasks/main.yml
```diff
@@ -115,3 +115,9 @@
     ca_cert: "{{ os_migrate_conversion_ca_cert|default(omit) }}"
     client_cert: "{{ os_migrate_conversion_client_cert|default(omit) }}"
     client_key: "{{ os_migrate_conversion_client_key|default(omit) }}"
+
+- name: install os_migrate conversion host content
+  hosts: os_migrate_conversion_host_name
```
ccamacho, line 120: "{{ os_migrate_conversion_host_name }}"

rbrady: thanks.  fixed in latest patch

os_migrate/playbooks/conversion_host_content.yml
```diff
@@ -0,0 +1,3 @@
+- hosts: conversion_hosts
```
rbrady, line 1: I don't think this is correct yet.  It's telling me the conversion host OS is Fedora.  According to the os_migration notes in gitlab, the conversion host os should be CentOS based.  Am I targeting the conversion host properly by using the hosts: block on the playbook and using the group name created in the conversion_host role?

os_migrate/roles/conversion_host/tasks/main.yml
```diff
@@ -115,3 +115,8 @@
     ca_cert: "{{ os_migrate_conversion_ca_cert|default(omit) }}"
     client_cert: "{{ os_migrate_conversion_client_cert|default(omit) }}"
     client_key: "{{ os_migrate_conversion_client_key|default(omit) }}"
+
+- name: install os_migrate conversion host content
+  include_role:
+    name: conversion_host_content
+  when: os_migrate_conversion_host_content_install
```
jistr, line 122: This is not the correct location to include the role - it would run it on the migrator host, not on conversion hosts. We should add this `include_role` at the end of `deploy_conversion_hosts.yml` playbook, in a similar manner like this:

https://github.com/os-migrate/os-migrate/blob/083d5796b5514aba335a08786cdc664644c813ea/os_migrate/playbooks/deploy_conversion_hosts.yml#L49-L54

This will give us the ability to specify which hosts it should run on.

os_migrate/playbooks/conversion_host_content.yml
jistr: Yes but we'd also need to run the conversion host inventory tasks to have the `conversion_hosts` group available:

https://github.com/os-migrate/os-migrate/blob/master/os_migrate/roles/conversion_host/tasks/conv_host_inventory.yml

I think we should remove this playbook (i don't think users would typically want to try and install the content without deploying the conversion hosts first) and add this role conditionally into conversion host deployment playbook as i've suggested in the other comment. There we already have the `conversion_hosts` group available because the tasks which add it were executed earlier in the playbook.

ccamacho:
This is the
  CentOS 8 image name: CentOS-8-GenericCloud-8.2.2004-20200611.2.x86_64.qcow2
  RHEL 8.2 image: rhel-8.2-x86_64-kvm.qcow2

To start using them, just change in this PR the line:
https://github.com/os-migrate/os-migrate/blob/master/tests/e2e/scenario_variables.yml#L12   and
https://github.com/os-migrate/os-migrate/blob/master/tests/e2e/scenario_variables.yml#L16 to CentOS-8-GenericCloud-8.2.2004-20200611.2.x86_64.qcow2 and when we trigger the 13-13 job it should run.

jistr:
I would say let's try switching the CI to the CentOS 8 image in this patch, and let's merge as soon as it passes. That way we get rid of the hand-built image and we can progress towards 10->13 CI.

rbrady:
As soon as I’m done getting RHEL to work, I will switch this patch to using CentOS.

os_migrate/roles/conversion_host_content/tasks/main.yml
```diff
@@ -0,0 +1,18 @@
+- name: display host os
+  debug: msg="Conversion Host OS is {{ ansible_distribution }}."
+
+- name: install epel release for CentOS
+  yum:
```
ccamacho, line 5: We will probably need to add a become: True in any install task
Then main.yml should call both centos and rhel tasks

rbrady: The become added to the task within the role doesn't work: https://gitlab.cee.redhat.com/os-migrate/bot/-/jobs/1839641

Adding the become at the task where the role is imported doesn't work either: https://gitlab.cee.redhat.com/os-migrate/bot/-/jobs/1839872

My guess is that I need to figure out what user has sudo on the image being used and add that to the end to end testing cfg

ccamacho:
I managed to find the missing parts to make this PR to work, please @rbrady when having some time include the missing bits from #240 here. Thanks!!

os_migrate/roles/conversion_host/tasks/main.yml
ccamacho: Fixed

jistr:
This commit is included in #240, Ryan's commit authorship is preserved there and Carlos' tweaks on top make the CI pass, so we'll merge.

jistr:
Merged as part of #240

PR #247, opened 2020-07-28
Adds Import and Export Roles for Nova Flavors

rbrady:
This patch adds import and export roles, supporting files and tests
for migrating Nova Flavors between clouds.

### DIFF
--- a/os_migrate/playbooks/export_flavors.yml
+++ b/os_migrate/playbooks/export_flavors.yml
@@ -1,0 +1,3 @@
+- hosts: migrator
+  roles:
+    - os_migrate.os_migrate.export_flavors
--- a/os_migrate/playbooks/import_flavors.yml
+++ b/os_migrate/playbooks/import_flavors.yml
@@ -1,0 +1,3 @@
+- hosts: migrator
+  roles:
+    - os_migrate.os_migrate.import_flavors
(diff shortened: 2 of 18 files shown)


ccamacho:
Hi Ryan! This is awesome! Pretty close I think. I remember having this issue and what I did is to install devstack with the same options locally in another dev box, and used the Nova config files in the job. It passed that part at least on my tests back then.

jistr:
What about trying to re-enable n-cond now? It should create the cell1.conf:

https://github.com/openstack/devstack/blob/40879683cc9719b5791ce6755f747958c9ef9cd3/lib/nova#L520-L524

https://github.com/openstack/devstack/blob/40879683cc9719b5791ce6755f747958c9ef9cd3/lib/nova#L733-L736

ccamacho:
Yeah, what I didn't have before was the dummy rabbit service. I think this time is closer than what I managed to move it before.

jistr:
Looking again at the `lib/nova` in devstack, do we need to add "placement" or "placement-client" or both into the enabled services list? I'm not sure where the endpoint is created, but there is some extra config going on here https://github.com/openstack/devstack/blob/master/lib/nova#L540

.github/workflows/consistency-functional.yml
```diff
@@ -168,22 +104,40 @@ jobs:
           cd devstack
           cat << EOF > local.conf
```
SeanMooney, line 105: EOF should be 'EOF' to prevent the varibales being substituted

PR #252, opened 2020-08-12
Adds compute services to CI

rbrady:
This patch adds Nova compute services to the CI workflow.


rbrady:
Going in a different direction.  See #253

PR #253, opened 2020-08-12
Adds Nova to CI Stack

rbrady:


rbrady:
Removed the DNM and changed the commit msg.

PR #257, opened 2020-08-28
Adds Roles For Nova Keypair Import and Export

rbrady:
This patch adds the roles and associated code required to export
and import Nova keypairs.


jistr:
We could add e2e test for keypairs too, but not a blocker, can be done separately.

jistr (approved):

PR #262, opened 2020-09-03
docs refactor

ccamacho:


ccamacho:
I'm adding a full check here because I'm updating almost 100 files, even if the change should not affect any functionality in os-migrate

jistr (approved):

PR #268, opened 2020-09-15
Add Tenant(Project) Roles

rbrady:
This patch adds the roles and supporting code to import and export
OpenStack projects.


ccamacho (commented):
When you have some time, for the docs is required to add references
here:
  - https://github.com/os-migrate/os-migrate/tree/main/docs/src/roles
and here:
  - https://github.com/os-migrate/os-migrate/tree/main/docs/src/modules

rbrady:
Thanks for the heads up on docs, I need to add those steps to the contributing doc.

os_migrate/roles/export_projects/README.md
```diff
@@ -0,0 +1,2 @@
+Please refer to
```
ccamacho, line 1: This should point similar to: https://github.com/os-migrate/os-migrate/pull/270/files#diff-c3d7c71d6d79498900947637e86723ddL2

ccamacho (commented):
For the current CI error, the modules should be included here:
os_migrate/tests/sanity/ignore-2.9.txt

os_migrate/plugins/module_utils/project.py
```diff
@@ -0,0 +1,36 @@
...
+    import const, resource
+
+
+class Project(resource.Resource):
+    resource_type = const.RES_TYPE_PROJECT
+    sdk_class = openstack.identity.v3.project.Project
+
+    info_from_sdk = [
+        'domain_id',
+        'is_enabled',
```
jistr, line 16: `is_enabled` should be a param if the create method supports it.

jistr (changes requested):

tests/func/run/all.yml
```diff
@@ -69,3 +69,11 @@
...
         - test_keypair
   tags: always
+
+- name: Include project tasks
+  include_tasks: project.yml
+  args:
+    apply:
+      tags:
+        - test_project
+  tags: always
```
jistr, line 79: Looked at the test results now, yeah this won't run. We'll need to do the split of Makefile targets and auth files and test playbooks between admin and tenant variants. I think we can mirror what was done in e2e tests and it should work.

jistr (changes requested):

rbrady:
still investigating why I get two instances of the seeded project in the test.

rbrady:
The issue appears to be where the import_project is run in the idempotent test, the project "osm_project" is imported successfully.  The credentials for the test_as_admin are same for src and dest tenants, and the creation of two projects with the same name is allowed in the openstack api.  This causes a case that results in two projects named "osm_project" in the same tenant.  I'm not sure if the best case here is to either create and use a separate admin user for the admin tests (admin-alt) or add the code to artificially restrict creating two projects with the same name.

jistr (commented):
Export idempotence logic seems to be the issue here, it relies on `id` property. We should add a special case for Project resources, for which it should rely on the tuple (domain_id, name), as Projects don't have IDs but they are uniquely named within a domain. The special case can be inserted here:

https://github.com/os-migrate/os-migrate/blob/main/os_migrate/plugins/module_utils/serialization.py#L61-L65

rbrady:
Added the special case in serialization.py and updated the func tests as needed.

jistr (approved):
LGTM. I don't think this would influence e2e tests but it's a large patch so i'm adding a CI run just to make sure.

PR #269, opened 2020-09-16
export/import users

ccamacho:


os_migrate/plugins/module_utils/const.py
```diff
@@ -15,4 +15,5 @@
 RES_TYPE_SECURITYGROUP = 'openstack.network.SecurityGroup'
 RES_TYPE_SERVER = 'openstack.compute.Server'
 RES_TYPE_SUBNET = 'openstack.subnet.Subnet'
+RES_TYPE_USER = 'openstack.subnet.User'
```
rbrady, line 18: is ‘subnet’ a typo here since you’re using the identity.v3.user in user.py?

rbrady (changes requested):

os_migrate/plugins/module_utils/const.py
ccamacho: Yes sorry a typo, thanks for the check

os_migrate/plugins/module_utils/user.py
```diff
@@ -0,0 +1,34 @@
...
+    sdk_class = openstack.identity.v3.user.User
+
+    info_from_sdk = [
+        'domain_id',
+        'is_enabled',
+    ]
+
+    params_from_sdk = [
+        'name',
+        'password',
```
jistr, line 21: This is surprising - can we really migrate the password? I think this probably doesn't work and should be in _info, no?

And on the contrary, there are additional things that we probably do want to migrate - is_enabled, description, maybe default_project_name (via reference id/name mechanisms) too?

Here are the docs: https://docs.openstack.org/openstacksdk/latest/user/resources/identity/v3/user.html#openstack.identity.v3.user.User 

Domain_id is probably correctly in _info. Trying to save a domain_name reference would probably break it for Keystone v2, and i think the users/projects will anyway be created in the domain which is provided in admin credentials, if any.

scripts/linters.sh
```diff
@@ -52,3 +52,29 @@ find ./ \
      -and -not -wholename "./tests/func/tmp/*" \
      -and -name "*.yml" -print0 \
     | xargs -0 ansible-lint
+
```
jistr, line 55: Not a blocker, but next time we should be putting such things into their own commits. (Not necessarily their own PRs, but at least commit separation is good.)

jistr (changes requested):

os_migrate/plugins/module_utils/user.py
jistr: Re password - it probably doesn't make sense to have it in _info either, i doubt the API would return the acutal password or even the stored hash, so we can probably drop it altogether?

scripts/linters.sh
ccamacho: Moving this after is ok before merge

ccamacho: done

os_migrate/plugins/module_utils/user.py
ccamacho: I think we should leave the password and iterate over, looks like a bit useless to migrate 1000s of users that need to restart their password? Everything else is fixed.

os_migrate/plugins/module_utils/user.py
```diff
@@ -0,0 +1,60 @@
...
+
+    sdk_params_from_refs = [
+        'default_project_id',
+    ]
+
+    @staticmethod
+    def _refs_from_sdk(conn, sdk_res):
+        refs = {}
+        refs['default_project_ref'] = reference.user_ref(
+            conn, sdk_res['default_project_id'])
```
jistr, line 39: We can't use user_ref on a project id. We're looking up a user and passing it a project id - so that will not find any user :). We need to use project_ref function here.

os_migrate/plugins/module_utils/user.py
```diff
@@ -0,0 +1,60 @@
...
+        refs['default_project_id'] = sdk_res['default_project_id']
+        return refs
+
+    def _refs_from_ser(self, conn, filters=None):
+        refs = {}
+        refs['default_project_ref'] = self.params()['default_project_ref']
+        refs['default_project_id'] = reference.user_id(
+            conn, self.params()['security_group_ref'])
```
jistr, line 47: Same here, we need to use project_id function. Also, there's copypasta with "security_group_ref" here.

os_migrate/plugins/module_utils/reference.py
```diff
@@ -249,6 +249,26 @@ def subnet_pool_ref(conn, id_, required=True):
     return _fetch_ref(conn, conn.network.find_subnet_pool, id_, required)
 
 
+def user_id(conn, ref, required=True):
```
jistr, line 252: We probably don't need these methods yet but they look good so feel free to keep them.

jistr (changes requested):

os_migrate/roles/export_users/tasks/main.yml
```diff
@@ -0,0 +1,37 @@
...
+        | os_migrate.os_migrate.stringfilter(os_migrate_users_filter,
+                                             attribute='id') ) }}"
+
+- name: export users
+  os_migrate.os_migrate.export_user:
+    auth: "{{ os_migrate_src_auth }}"
+    auth_type: "{{ os_migrate_src_auth_type|default(omit) }}"
+    region_name: "{{ os_migrate_src_region_name|default(omit) }}"
+    path: "{{ os_migrate_data_dir }}/users.yml"
+    name: "{{ item['name'] }}"
```
jistr, line 32: Should be item['id'] ?

jistr (changes requested):
(inline)

os_migrate/roles/export_users/tasks/main.yml
jistr: In case of users it might not matter, OpenStack allows duplicate naming on many resources but i doubt it would allow it on users (?), but still it would be nice to keep consistent with how we look up resources elsewhere.

ccamacho: fixed, it was one of the changes I did for trying to see why I wasnt able to find the user

os_migrate/plugins/module_utils/reference.py
```diff
@@ -249,6 +249,26 @@ def subnet_pool_ref(conn, id_, required=True):
...
+    """Fetch ID of a Project identified by reference dict `ref`. Use
+    OpenStack SDK connection `conn` to fetch the info. If `required`,
+    ensure the fetch is successful.
+    Returns: the ID, or None if not found and not `required`
+    Raises: openstack's ResourceNotFound when `required` but not found
+    """
+    return _fetch_id(conn, conn.identity.find_project, ref, required)
+
+
+def project_ref(conn, id_, required=True, get_project_info=False):
```
jistr, line 262: We should remove that parameter from this function signature and always pass False below. See flavor_ref on line 225:

https://github.com/os-migrate/os-migrate/pull/269/files#diff-b38d550e0c19cfb70943298e7af3a33fR225

jistr (changes requested):

jistr (approved):

PR #271, opened 2020-09-21
include admin tasks

ccamacho:


tests/e2e/admin/clean.yml
```diff
@@ -0,0 +1,231 @@
+- name: remove osm_server
```
jistr, line 1: I don't think we want to test server migration from admin project to admin project?

We probably should discuss what makes sense testing with what credentials. IMO:

* admin user + admin project -> resources like flavors, users, projects. That's this job you are editing. We should not confuse this with "migrate tenant resources as admin"

* tenant user + tenant project -> what we already test in our main jobs

* admin user + tenant project -> this is the "migrate tenant resources as admin" which requires the role setup that you documented earlier. Probably the same resources should be tested here as in the tenant user + tenant project jobs.

I guess this is quite confusing so we should probably discuss this on a call, what's possible, what makes sense to test, how to document it for users.

jistr (changes requested):

PR #283, opened 2020-10-12
Resource check

rbrady:


jistr (changes requested):
Does this need a rebase on top of the latest main branch? The diff looks like the PR is introducing export/import for projects, but that was already merged earlier. So there's a lot of irrelevant code in the review diff.

rbrady:
> Does this need a rebase on top of the latest main branch?

Rebased.

os_migrate/plugins/module_utils/serialization.py
```diff
@@ -16,7 +16,7 @@ def new_resources_file_struct():
     return data
 
 
-def add_or_replace_resource(resources, resource_or_data):
+def add_or_replace_resource_old(resources, resource_or_data):
```
jistr, line 19: We can delete this `_old` function now, right? The refactored function looks good, and the rest of the patch too. :+1:

jistr (changes requested):

jistr (approved):
Thanks!

PR #284, opened 2020-10-14
Reuse container images if possible

ccamacho:


ccamacho:
I need to check how to inject the current code in the container image prior to running the job.

ccamacho:
The previous build, created the container image as we were updating the Makefile, in this run, I omitted the file and it reused the latest available.

jistr (approved):
Looks good, thanks!

PR #293, opened 2020-10-21
Include changelog in docs

ccamacho:


jistr (changes requested):
This looks good to me in general, i think we should discuss how we want to use it. I'm +1 for the new/chg/fix distinction, but i'm not that eager to do the usr/dev/test distinction.

I'm giving -1 for now though because i think we should use the script from pypi (it seems to be released there) rather than copy it into our repo. Would that be possible?

ccamacho:
Agreed on using the pypi if we dont need any other change anymore, but for example, there are some bugs in sphinx that will make us do changes in the script, for instance:

- I want to link from the description of each release, the docs changelog permalink (right now they are id1, id2 so we can not link them).
- The library has a lot of time without being released (2 years without any update) and we might hit python bugs in the near future
- The library needs multiple lint fixes.
- If we want to sort the changes in a different way for reporting what we have we will need to make local changes.

I think having the library locally in the scripts folder will give us a lot of flexibility if we need to make changes. The library supports multiple formats (which we can remove) and get only with the rst feature and only the bits we need. We can discuss it next week on the squad call.

jistr:
> Agreed on using the pypi if we dont need any other change anymore, but for example, there are some bugs in sphinx that will make us do changes in the script, for instance:
> 
>     * I want to link from the description of each release, the docs changelog permalink (right now they are id1, id2 so we can not link them).

I don't fully understand what you mean here, let's discuss on a the call next week?

> 
>     * The library has a lot of time without being released (2 years without any update) and we might hit python bugs in the near future

That makes me wonder whether we should look at something else even... It might be better if we can just use some maintained changelog generator rather than start maintaining our own? I see some pull requests which were opened a long time ago and didn't receive any feedback:

https://github.com/vaab/gitchangelog/pulls

> 
>     * The library needs multiple lint fixes.

That wouldn't be an issue if we used it from pypi though.

> 
>     * If we want to sort the changes in a different way for reporting what we have we will need to make local changes.

Yea, on one hand i agree, on the other hand my gut feeling is that we shouldn't spend much time maintaining a perfect changelog, maybe better to have a good enough changelog generated by a maintained tool, and focus our continuous dev time on OS Migrate improvements. I'm not 100% sold either way here, just describing my thought process and why i'm not comfortable giving this a +1 review right now.

> 
> 
> I think having the library locally in the scripts folder will give us a lot of flexibility if we need to make changes. The library supports multiple formats (which we can remove) and get only with the rst feature and only the bits we need. We can discuss it next week on the squad call.

Yea that's a fair point, using an in-tree script has its benefits too. I'd say i'm on the fence, let's discuss on the call also with @rbrady .

jistr:
Also, if we really pull in the script, we would need to pull in its license file too. We'd need to adhere to the license terms there.

jistr:
On a quick bunch of searches i'm not finding that many good alternatives... But i'm still thinking:

* Let's use it from PyPi for now if at all possible (we'd discuss on the call what are the issues?).

* If/when we discover PyPi isn't usable, we pull the script in and edit, but we gotta watch out for the license terms. I'd probably create `vendor/gitchangelog` dir. In general, code which got into the repo this way would best be separated into a `vendor` dir i think.

ccamacho:
> This looks good to me in general, i think we should discuss how we want to use it. I'm +1 for the new/chg/fix distinction, but i'm not that eager to do the usr/dev/test distinction.
> 
> I'm giving -1 for now though because i think we should use the script from pypi (it seems to be released there) rather than copy it into our repo. Would that be possible?

Yeah :) I still think we need the library, but, it's something we can definitely add in the future as a follow-up patch, to unblock this I already removed the python script and installed it by pip.

jistr (approved):
Thanks!

PR #295, opened 2020-10-22
Fix: Revert to conditional empty variable check

ccamacho:
Im following now the same approach we have in another
variables check in the same makefile.
Even if locally the check works in the job we have
errors like: if [[ EUSE_TOOLBOX -eq "0" ]]; then \
Notice that EUSE instead or REUSE

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -151,8 +151,7 @@ test-unit: reinstall
 # TOOLBOX
 
 toolbox-build:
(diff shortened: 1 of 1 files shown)


ccamacho:
We were not catching this in the CI jobs because we always set the variable TOOLBOX_REUSE then we catch it when pushing the latest image to the image registry.

jistr (approved):

PR #298, opened 2020-11-05
Take info about server ports from Neutron

jistr:
Previously we took it from Nova but that does not contain all the
needed info for advanced use cases. Now we export info from Neutron
which should allow us to export more port properties. The creation of
ports during server import is still kept simple for now, via Nova.


rbrady (approved):
LGTM

jistr:
Thanks for the review :)

PR #299, opened 2020-11-10
Workload Source Check

rbrady:
During the workload import, the code was checking the workload status
via the export file.  If a server had been turned on or moved into
another state other than "SHUTOFF" since the export occured, the
workload import would fail.  This patch addresses that by checking
the source cloud via the openstacksdk to ensure the workload status
is "SHUTOFF" before attempting to migrate it.

### DIFF
--- a/os_migrate/plugins/modules/import_workload_source_check.py
+++ b/os_migrate/plugins/modules/import_workload_source_check.py
@@ -1,0 +1,139 @@
+#!/usr/bin/python
+
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+ANSIBLE_METADATA = {
+    'metadata_version': '1.1',
+    'status': ['preview'],
+    'supported_by': 'community'
+}
+
+DOCUMENTATION = '''
+---
+module: import_workload_source_check
+
+short_description: Export OpenStack instance information
+
+extends_documentation_fragment: openstack
+
+version_added: "2.9"
+
+author: "OpenStack tenant migration tools (@os-migrate)"
+
+description:
+  - "Check OpenStack workload in source cloud"
+
+options:
+  auth:
+    description:
+      - Dictionary with parameters for chosen auth type.
+    required: true
+    type: dict
+  auth_type:
+    description:
+      - Auth type plugin for OpenStack. Can be omitted if using password authentication.
+    required: false
+    type: str
+  validate_certs:
+    description:
+      - Validate HTTPS certificates when logging in to OpenStack.
+    required: false
+    type: bool
+  region_name:
+    description:
+      - OpenStack region name. Can be omitted if using default region.
+    required: false
+    type: str
+  name:
+    description:
+      - Name (or ID) of an instance to check.
+    required: true
+    type: str
+  availability_zone:
+    description:
+      - Availability zone.
+    required: false
+    type: str
+  cloud:
+    description:
+      - Ignored. Present for backwards compatibility.
+    required: false
+    type: raw
+'''
+
+EXAMPLES = '''
+- name: ensure workload in source cloud is ready to continue
+  os_migrate.os_migrate.import_workload_source_check:
+    auth: "{{ os_migrate_src_auth }}"
+    auth_type: "{{ os_migrate_src_auth_type|default(omit) }}"
+    region_name: "{{ os_migrate_src_region_name|default(omit) }}"
+    validate_certs: "{{ os_migrate_src_validate_certs|default(omit) }}"
+    ca_cert: "{{ os_migrate_src_ca_cert|default(omit) }}"
+    client_cert: "{{ os_migrate_src_client_cert|default(omit) }}"
+    client_key: "{{ os_migrate_src_client_key|default(omit) }}"
+    name: migration-vm
+  when: prelim.changed
+'''
+
+RETURN = '''
+'''
+
+from ansible.module_utils.basic import AnsibleModule
+from ansible.module_utils.openstack \
+    import openstack_full_argument_spec, openstack_cloud_from_module
+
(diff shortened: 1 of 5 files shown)


rbrady:
Needs update to the tests...coming soon.

os_migrate/roles/import_workloads/tasks/workload.yml
```diff
@@ -21,6 +21,17 @@
...
+  - name: perform required checks to ensure source workload is ready
+    os_migrate.os_migrate.import_workload_source_check:
+      auth: "{{ os_migrate_src_auth }}"
+      auth_type: "{{ os_migrate_src_auth_type|default(omit) }}"
+      region_name: "{{ os_migrate_src_region_name|default(omit) }}"
+      validate_certs: "{{ os_migrate_src_validate_certs|default(omit) }}"
+      ca_cert: "{{ os_migrate_src_ca_cert|default(omit) }}"
+      client_cert: "{{ os_migrate_src_client_cert|default(omit) }}"
+      client_key: "{{ os_migrate_src_client_key|default(omit) }}"
+      name: "{{ prelim.server_name }}"
```
jistr, line 33: We should be looking up by ID, all the workload migration modules use IDs for the source server, the server_name is just for display purposes and logs. E.g. https://github.com/os-migrate/os-migrate/blob/f3d6dd70ad65c9ce403e76a76b4c2f18714423ab/os_migrate/plugins/modules/import_workload_export_volumes.py#L507 .

We have some options:

* use `name: "{{ item['_info']['id'] }}"` here, no other changes necessary i think,

* use `name: "{{ prelim.src_server_id }}"` here and add `result['src_server_id'] = info['id']` into the prelim module: https://github.com/os-migrate/os-migrate/blob/f3d6dd70ad65c9ce403e76a76b4c2f18714423ab/os_migrate/plugins/modules/import_workload_export_volumes.py#L507. This might be nicer as when running Ansible with -v, we get this ID printed out into the logs too.

os_migrate/plugins/modules/import_workload_source_check.py
```diff
@@ -0,0 +1,138 @@
...
+    #: The state this server is in. Valid values include ``ACTIVE``,
+    #: ``BUILDING``, ``DELETED``, ``ERROR``, ``HARD_REBOOT``, ``PASSWORD``,
+    #: ``PAUSED``, ``REBOOT``, ``REBUILD``, ``RESCUED``, ``RESIZED``,
+    #: ``REVERT_RESIZE``, ``SHUTOFF``, ``SOFT_DELETED``, ``STOPPED``,
+    #: ``SUSPENDED``, ``UNKNOWN``, or ``VERIFY_RESIZE``.
+    # Make sure source instance is shutdown before proceeding.
+    if srv.info['status'] != 'SHUTOFF':
+        name = module.params['name']
+        msg = 'Skipping instance {} because it is not in state SHUTOFF!'
+        module.exit_json(skipped=True, skip_reason=msg.format(name), **result)
```
jistr, line 128: This is going to print a message but otherwise the migration will not be affected. One option to fix this would be to register the result of this module, and in `workload.yml` all the conditions with `prelim.changed` would change to `prelim.changed and source_check.changed`.

But i'd even argue we should fail in this module instead of skipping. If the user has an instance which passed the `os_migrate_workloads_filter`, then they probably want to migrate it, and would prefer an actual failure if that can't happen, rather than silently skipping and having the playbook succeed. Then no further changes to `workloads.yml` are necessary for this. The `prelim` module will still be used to auto-skip the conversion host from trying to migrate itself, should it pass the filter by accident.

jistr (changes requested):

os_migrate/roles/import_workloads/tasks/workload.yml
jistr: (I edited the above comment - one of the source links was wrong, should be good now.)

jistr: I noticed one more thing, we need to give this `when: prelim.changed` in case the prelim module decided to skip the migration because it was trying to migrate the conversion host itself.

rbrady:
Implemented review feedback

os_migrate/plugins/modules/import_workload_source_check.py
```diff
@@ -0,0 +1,139 @@
...
+
+    # Status Check
+    #: The state this server is in. Valid values include ``ACTIVE``,
+    #: ``BUILDING``, ``DELETED``, ``ERROR``, ``HARD_REBOOT``, ``PASSWORD``,
+    #: ``PAUSED``, ``REBOOT``, ``REBUILD``, ``RESCUED``, ``RESIZED``,
+    #: ``REVERT_RESIZE``, ``SHUTOFF``, ``SOFT_DELETED``, ``STOPPED``,
+    #: ``SUSPENDED``, ``UNKNOWN``, or ``VERIFY_RESIZE``.
+    # Make sure source instance is shutdown before proceeding.
+    if srv.info['status'] != 'SHUTOFF':
+        msg = 'Skipping instance {} because it is not in state SHUTOFF!'
```
jistr, line 128: Should we reword the message now that we're not skipping? Something like 'Instance {} cannot be migrated because it is not in state SHUTOFF!' perhaps?

jistr (commented):
Just a message nit

os_migrate/plugins/modules/import_workload_source_check.py
rbrady: Changed

jistr (approved):
Looks like the last branch push was done in the middle of e2e CI run, i'll retrigger just in case, but all LGTM.

jistr:
This probably just needs rebasing on top of latest main branch, i'll try doing that.

PR #300, opened 2020-11-11
Dev: Docs minor improvement

ccamacho:
This patch adds a docs reference
for the commit's style convention.

Please enter the commit message for your changes. Lines starting

### DIFF
--- a/docs/src/devel/contributing.rst
+++ b/docs/src/devel/contributing.rst
@@ -165,7 +165,7 @@ associated export role, reading the data file and then calling the
 module you created for import.
 
 
-Commit messages
+Commit Messages
 ---------------
 
 For every pull request we request contributors to be compliant with the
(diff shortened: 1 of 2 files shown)


rbrady (approved):
LGTM.  One suggestion is to change the non-title case title from "Commit's information" to "Commit Messages".

docs/src/devel/contributing.rst
```diff
@@ -164,6 +164,51 @@ data. A common pattern is validating the data file created by the
 associated export role, reading the data file and then calling the
 module you created for import.
 
+
+Commit's messages
```
rbrady, line 168: s/Commit's messages/Commit Messages/

There should not be possessiveness of the messages to the commit. :)  Also, title case.

rbrady (changes requested):

docs/src/devel/contributing.rst
```diff
@@ -164,6 +164,51 @@ data. A common pattern is validating the data file created by the
...
+    |       +-> Summary in present tense.
+    |
+    +-------> Type: [Nn]ew, [cC]hg, [fF]ix.
+
+    <body> ----> The commit's body.
+
+Accepted types:
+
+- `new` or `New`: newly implemented features
+- `chg` or `Chg`: changes in the CI automation, documentation or any other change not presented as a new feature
```
jistr, line 191: I'd say CI automation should have a 'dev:' commit as it's not end-user facing. But this is just a nit. I think we can merge and tweak subsequently as we review docs.

jistr (approved):

docs/src/devel/contributing.rst
ccamacho: Cool we just need to adjust the changelog config when we update the docs so they are consistent, this patch covers only what we already have

PR #301, opened 2020-11-11
New: Adds Logic to Unsubscribe RHEL Conversion Hosts

rbrady:
This patch adds logic to unsubscribe systems from RHSM when using
RHEL as the OS for a conversion host.  The unsubscribe is called
for both the source and destination conversion hosts just before
they are deleted.

### DIFF
--- a/os_migrate/playbooks/delete_conversion_hosts.yml
+++ b/os_migrate/playbooks/delete_conversion_hosts.yml
@@ -2,6 +2,24 @@
   tasks:
     - include_role:
         name: os_migrate.os_migrate.conversion_host
+        tasks_from: conv_hosts_inventory.yml
+
+- hosts: os_migrate_conv_src
+  tasks:
+    - include_role:
+        name: os_migrate.os_migrate.conversion_host
+        tasks_from: unregister.yml
+
+- hosts: os_migrate_conv_dst
+  tasks:
+    - include_role:
+        name: os_migrate.os_migrate.conversion_host
+        tasks_from: unregister.yml
+
+- hosts: migrator
+  tasks:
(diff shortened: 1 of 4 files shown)


os_migrate/playbooks/delete_conversion_hosts.yml
```diff
@@ -1,3 +1,15 @@
+- hosts: os_migrate_conv_src
```
jistr, line 1: The src/dst hosts won't be in the inventory at this point. We'll need to first include the conversion host role with `tasks_from: conv_hosts_inventory.yml` from the migration host, that's what creates the conversion host inventory entries.

(I guess this wasn't tested? It would be nice to test this at least manually before merging.)

jistr (changes requested):

os_migrate/playbooks/delete_conversion_hosts.yml
rbrady: Not tested yet.  Wanted to get a review (glad I did).  My assumption is that this would only be testable via e2e, so that was my plan on where to test.

jistr:
I'm trying to test this locally with my credentials, so far i'm hitting PEBKAC issues, but i'm progressing :).

jistr:
I added some resilience to the unregistration (if any conversion hosts are missing or unreachable via ssh, don't fail and continue with ensuring both hosts are deleted), and tested manually with RHEL 8.3, WFM.

ccamacho (approved):

odyssey4me (approved):
LGTM

jistr (approved):

PR #305, opened 2020-11-20
Chg: User documentation improvements

jistr:
Improve clarity of user docs in various spots and add more workload
migration info to the walkthrough doc.


jistr:
I'm not sure if the

PR #306, opened 2020-11-20
Chg: User documentation improvements

jistr:
Improve clarity of user docs in various spots and add more workload
migration info to the walkthrough doc.


jistr:
Still jobs stuck in expected, will resubmit again.

PR #308, opened 2020-11-20
Dev: fix ci for file-changes-action

ccamacho:
There is a bug https://github.com/trilom/file-changes-action/issues/104
solved in latest version.

Currently the CI is broken.


jistr (changes requested):
Let's please fix the commit message - CI breakages and fixes is something that end users don't care about, so let's tag it Dev so that it doesn't show up in changelog.

jistr (approved):
Thanks!

PR #316, opened 2020-11-28
Dev: Include role documentation in sphinx template

ccamacho:
This patch renders the documentation including
the initial README.md file from each role.


jistr (approved):
Great!

Issue #317, opened 2020-11-30
Func CI broken because of devstack?


jistr:
I just noticed this on our recently merged patches, it may not be intermittent.

https://github.com/os-migrate/os-migrate/runs/1474395074?check_suite_focus=true#step:16:4323 

```
 Using python 3.6 to install /opt/stack/keystone
+ inc/python:pip_install:193               :   sudo -H LC_ALL=en_US.UTF-8 SETUPTOOLS_USE_DISTUTILS=stdlib http_proxy= https_proxy= no_proxy= PIP_FIND_LINKS= SETUPTOOLS_SYS_PATH_TECHNIQUE=rewrite python3.6 -m pip install -c /opt/stack/requirements/upper-constraints.txt -e /opt/stack/keystone
DEPRECATION: Constraints are only allowed to take the form of a package name and a version specifier. Other forms were originally permitted as an accident of the implementation, but were undocumented. The new implementation of the resolver no longer supports these forms. A possible replacement is replacing the constraint with a requirement.. You can find discussion regarding this at https://github.com/pypa/pip/issues/8210.
ERROR: Links are not allowed as constraints
```

jistr:
I am suspecting this might be related to a new pip release with stricter resolver:

https://pypi.org/project/pip/20.3/

Released just hours ago, so it could be the trigger.

PR #320, opened 2020-12-01
Dev: Revert "Fix functional CI by pinning pip version"

jistr:
This reverts commit c0ae23709ef96d6426fc9219e2dc56dd71c5f56a. It was a
temporary measure to fix CI broken by Devstack, which was in turn
broken by a new pip release.

Context:

https://github.com/os-migrate/os-migrate/issues/317
https://github.com/os-migrate/os-migrate/pull/318

### DIFF
--- a/.github/workflows/consistency-functional.yml
+++ b/.github/workflows/consistency-functional.yml
@@ -122,9 +122,7 @@ jobs:
           sudo apt install python3-httplib2 -y
           # We need to be sure we use the latest versions of
           # pip, virtualenv and setuptools
-          # FIXME: `git revert` the pip version constraint when devstack is fixed for new pip
-          #        https://github.com/os-migrate/os-migrate/issues/317
-          sudo python -m pip install --upgrade "pip<20.3"
+          sudo python -m pip install --upgrade pip
           sudo python -m pip install --upgrade virtualenv
           sudo python -m pip install --upgrade setuptools
       - name: Verify MySQL connection from container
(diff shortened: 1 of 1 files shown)


jistr:
Devstack has now implemented the same workaround we have here. Retrying the CI.

https://github.com/openstack/devstack/commit/7a3a7ce876a37376fe0dca7278e41a4f46867daa

.github/workflows/consistency-functional.yml
```diff
@@ -122,9 +122,7 @@ jobs:
           sudo apt install python3-httplib2 -y
           # We need to be sure we use the latest versions of
           # pip, virtualenv and setuptools
-          # FIXME: `git revert` the pip version constraint when devstack is fixed for new pip
-          #        https://github.com/os-migrate/os-migrate/issues/317
-          sudo python -m pip install --upgrade "pip<20.3"
+          sudo python -m pip install --upgrade pip
```
odyssey4me, line 125: Have you thought about using a test-requirements.txt file or something to pin these and then use something like https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/keeping-your-dependencies-updated-automatically to keep them up to date via PR whenever they change. That way the updates are tested before they're used.

odyssey4me (approved):

.github/workflows/consistency-functional.yml
jistr: We haven't, i think something like this would only make sense investing in if problems that it could prevent would happen regularly. Even in the case of this particular PR though, the solution you point out wouldn't help, because we needed to patch Devstack too.

PR #321, opened 2020-12-03
Dev: Fix CI Error

rbrady:
CI was failing on the step "Log into GitHub Container Registry".
This patch changed.....TBD


jistr:
Looks like this was a temporary problem, i just retriggered CI on the PRs and they went past this step without issues.

PR #327, opened 2020-12-10
New: Creation of Floating IPs on migrated workloads

jistr:
New: Creation of Floating IPs on migrated workloads
    
    A new migration parameter 'floating_ip_mode' is added into the
    workload serialization. The current possible values are 'auto' and
    'skip'.
    
    * 'auto' means that floating IP creation is attempted for the VM, if
      the source VM had floating IP(s). The floating IP info is already
      being exported with the workload.
    
      Currently we only support auto-assignment of the floating IP
      address. (Exact assignment is typically not allowed by policy for
      tenants, but this should be investigated too.)
    
      Each floating IP serialization has a 'floating_network_ref'
      reference, which specifies the network where the floating IP should
      be created. The 'fixed_ip_address' parameter controls which
      port/address will the floating IP attach to.
    
    * 'skip' means that even if the workload serialization does include
      serialized floating IP subresources, their creation in destination
      cloud will not be attempted.
    
    The 'floating_ip_mode' parameter defaults to 'auto'.

### DIFF
--- a/os_migrate/plugins/module_utils/server_floating_ip.py
+++ b/os_migrate/plugins/module_utils/server_floating_ip.py
@@ -2,9 +2,10 @@ from __future__ import (absolute_import, division, print_function)
 __metaclass__ = type
 
 import openstack
+import time
 
 from ansible_collections.os_migrate.os_migrate.plugins.module_utils \
-    import exc, const, resource
+    import exc, const, reference, resource
 
 
 def server_floating_ips(conn, server_ports):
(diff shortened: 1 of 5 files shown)


os_migrate/plugins/module_utils/server_floating_ip.py
```diff
@@ -31,30 +32,91 @@ class ServerFloatingIP(resource.Resource):
...
+        sdk_params = self._to_sdk_params(self._refs_from_ser(conn))
+        sdk_params['port_id'] = my_port['id']
+        return conn.network.create_ip(**sdk_params)
+
+    def _find_my_server_port(self, conn, sdk_srv, delay=5, retries=5):
+        params = self.params()
+        for try_number in range(retries):
+            srv_ports = conn.network.ports(device_id=sdk_srv['id'])
+            for port in srv_ports:
+                for addr in params['fixed_ip_address']:
```
jistr, line 86: BUG! There should be 'if addr' not 'for addr'.

jistr: We can't just return the first port, we need to check if the addresses match.

jistr: Fixed.

ccamacho (approved):

PR #328, opened 2020-12-10
Dev: Add time profiling into CI logs

jistr:
This should let us inspect how much time is consumed by individual
tasks.


jistr:
Looks like the package updating of conv hosts adds about 16 minutes to e2e CI times. We should parallelize the conversion host preparation.

ccamacho (approved):

PR #329, opened 2020-12-11
New: Parallel installation of conversion hosts

jistr:
The updating and installing of packages on conversion hosts can take
minutes. These actions were done in series until now (src then dst),
but now they're done in parallel, which should speed up execution of
the conversion host deployment playbook.


jistr:
All green, ready for review.

ccamacho (approved):

PR #330, opened 2020-12-14
docs: include information in all roles

ccamacho:


_skeleton_role_/README.md.j2
```diff
@@ -0,0 +1,5 @@
+The role {{ _role_name }}
```
jistr, line 1: And the skeleton role too. Should go into a "for devs" directory. We may rename the `scripts` directory later to something else, as it won't be just scripts anymore. But i wouldn't conflate adding role skeleton and renaming the scripts dir in a single PR. The rename has potential to break a bunch of things, so it should be a separate card/PR.

role-addition.yml
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env ansible-playbook
```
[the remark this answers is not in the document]
ccamacho: done I was just testing this and its addition in the docs.

_skeleton_role_/README.md.j2
ccamacho: done I was just testing this and its addition in the docs. I agree on changing the scripts name in an additional PR

rbrady (approved):

PR #332, opened 2021-01-08
Dev: use 'make docs' to generate docs in Github Actions

jistr:
Instead of using the modules which are specific to Github
Actions, let's use 'make docs' to build documentation, be it locally by a
contributor, or in CI.


jistr:
This is ready for review but we should merge it after #333 to make sure that the changelog is generated correctly.

ccamacho (approved):

PR #334, opened 2021-01-12
Dev: Fix meta template

ccamacho:
This patch fixes the skeleton template
for creating new roles.

### DIFF
--- a/scripts/_skeleton_role_/meta/main.yml.j2
+++ b/scripts/_skeleton_role_/meta/main.yml.j2
@@ -10,5 +10,12 @@ galaxy_info:
       versions:
         - 30
   galaxy_tags: ["os-migrate"]
+{% if 'export' in role_name %}
 dependencies:
   - role: os_migrate.os_migrate.prelude_src
+{% elif 'import' in role_name %}
+dependencies:
+  - role: os_migrate.os_migrate.prelude_dst
+{% else %}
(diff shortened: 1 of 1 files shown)


scripts/_skeleton_role_/meta/main.yml.j2
```diff
@@ -10,5 +10,12 @@ galaxy_info:
       versions:
         - 30
   galaxy_tags: ["os-migrate"]
+{% if 'export' in role_name %}
```
odyssey4me, line 13: Both conditionals could be done in this if statement to deduplicate a little.

jistr: There's 'prelude_src' vs. 'prelude_dst' difference in the dependency in each conditional though, so IMO it's good?

ccamacho: yup, they are different, I had the 'dependencies:' outside, but then I remembered that we might have different roles without any dependency

odyssey4me: Oh duh. OK never mind. 👍

odyssey4me (approved):

odyssey4me (approved):

odyssey4me (approved):

PR #335, opened 2021-01-14
Chg: Don't feed availability_zone_hints into requests when empty 

jistr:
Dev: Mechanism for skipping falsey params

Empty lists usually don't make sense to feed into the API requests as
parameters. Often the results of feeding an empty list vs. not
specifying the parameter can be the same, but it can matter in some
cases, e.g. if Neutron router availability zone extension is disabled,
the parameter must not be specified in the request.

--------

Chg: Don't feed availability_zone_hints into requests when empty

The availability_zone_hints parameter for routers and networks can
often be an empty list. Previously we've fed it as-is into the API
requests. However, if the extension for network/router availability
zones isn't enabled, this can cause the request to fail.

We now only use the availability_zone_hints parameter when its not
empty. This should have no effect on correctness, but it can prevent
false-alarm failures.

### DIFF
--- a/os_migrate/plugins/module_utils/resource.py
+++ b/os_migrate/plugins/module_utils/resource.py
@@ -33,10 +33,17 @@ class Resource():
     # Items of refs dict that are used as SDK call kwargs when
     # issuing a create/update REST API request.
     sdk_params_from_refs = []
-    # some parameters are allowed when creating a resource but not when
+    # Some parameters are allowed when creating a resource but not when
     # updating it.  This list of parameter names is purged from the parameter
     # list before calling update.
     readonly_sdk_params = []
+    # Some parameters get serialized as e.g. an empty array and it can
+    # matter whether they are fed into a create/update request or
+    # skipped altogether. Parameters listed here will not be included
+    # in sdk_params in case their value is falsey - false, empty
+    # array, empty dict etc. (Parameters with None value are skipped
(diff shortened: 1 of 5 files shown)


os_migrate/plugins/module_utils/resource.py
```diff
@@ -33,10 +33,17 @@ class Resource():
...
     # updating it.  This list of parameter names is purged from the parameter
     # list before calling update.
     readonly_sdk_params = []
+    # Some parameters get serialized as e.g. an empty array and it can
+    # matter whether they are fed into a create/update request or
+    # skipped altogether. Parameters listed here will not be included
+    # in sdk_params in case their value is falsey - false, empty
+    # array, empty dict etc. (Parameters with None value are skipped
+    # anyway and do not need to be listed here.)
+    skip_falsey_sdk_params = []
```
jistr, line 46: I think we're gradually getting to a point where instead of having extra lists like `readonly_sdk_params` and `skip_falsey_sdk_params`, we should probably refactor the parameter lists to be parameter dictionaries, and allow them to carry info about how the parameter should be treated. E.g.:

```python
params_from_sdk = {
    'availability_zone_hints': { 'skip_falsey': true },
    'some_normal_param': {},
    'some_readonly_param': { 'readonly': true },
}
```

I think that would be a separate refactoring task though.

ccamacho (approved):

PR #336, opened 2021-01-19
Chg: Pass through all RHSM parameters for conversion hosts

jistr:
Chg: Shorter names for RHSM variables

The RHSM variables are renamed from previous pattern of
e.g. 'os_migrate_conversion_host_content_rhsm_user' to
e.g. 'os_migrate_conversion_rhsm_user' which is shorter but should
still be descriptive enough and not risk a naming collision.

The 'user' variable is also renamed to 'username' to match the
parameter naming on the redhat_subscription module.

------------

Chg: Pass through all RHSM parameters for conversion hosts

All parameters supported by the redhat_subscription Ansible module are
now exposed as variables for the conversion host content installation.


ccamacho (approved):
LGTM

PR #338, opened 2021-01-19
Dev: Convert to fully qualified collection names

ccamacho:
Starting on Ansible 2.9 all tasks can use
the fully defined collection name the builtin
runtime is specified in:
https://github.com/ansible/ansible/blob/devel/lib/ansible/config/ansible_builtin_runtime.yml

### DIFF
--- a/.github/workflows/container-image-build.yml
+++ b/.github/workflows/container-image-build.yml
@@ -1,41 +1,40 @@
----
 name: container-image-build
 on:
   push:
     paths-ignore:
-      - 'doc/**'
+    - 'doc/**'
 jobs:
   push_to_reg:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - name: Install dependencies
-        run: |
-          sudo locale-gen en_US.UTF-8
-          sudo dpkg-reconfigure locales
-          sudo python -m pip install --upgrade pip
-          sudo pip install tox
-          sudo apt install software-properties-common -y
-          sudo add-apt-repository -y ppa:projectatomic/ppa
-          sudo apt update -y
-          sudo apt install build-essential findutils -y
-          sudo apt install podman -y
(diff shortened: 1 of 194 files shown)


.github/workflows/container-image-build.yml
```diff
@@ -1,41 +1,40 @@
----
 name: container-image-build
 on:
   push:
     paths-ignore:
-      - 'doc/**'
```
jistr, line 6: I'm not sure what prompted the whitespace changes but i'm -1 about this for a few reasons:

* What we already have seems to be the generally preferred way to write YAML, e.g. see the voting at: https://github.com/prettier/prettier/issues/4723.

* The voting there ^ matches my personal opinion as well. It matches a consistent approach of "when i'm putting something into a dictionary, it's 2 spaces more indented than the key". So e.g. a list of values like this:

    ```
    normal text

    - list
    - list

    dict1: 1
    dict2: 2
    ```

  Applying the same procedure to all of these we get:

    ```
    outer: >-
      normal text

    outer:
      - list
      - list

    outer:
      dict1: 1
      dict2: 2
    ```

    Also the clarity of the conventional approach is better IMO, consider these:

    ```
    outer:
      inner1: 1
      inner2: 2

    outer:
    - inner1: 1
    - inner2: 2

    outer:
      - inner1: 1
      - inner2: 2
    ```

    The indentation of the "inner" bits in the first two cases is the same, but in the first case it is just 'dict -> dict', while in the second case it is 'dict -> list -> dict', so it would make sense that the final indentation should be +2 spaces given it has +1 data structure nesting level. So IMO the 3rd case is the clearer way to represent these nested data structures, and we already use this approach.

* Doing any sweeping changes like this across the whole repo will make it really difficult to use `git blame` or the blame view on Github website.

jistr (changes requested):

Issue #350, opened 2021-01-25
All e2e CI jobs broken: Unrecognized attribute(s) ''remote_address_group_id'''


jistr:
TASK [os_migrate.os_migrate.conversion_host : allow ssh in os_migrate conversion security group] ***
Friday 22 January 2021  18:05:42 +0000 (0:00:01.281)       0:00:43.833 ******** 
fatal: [localhost]: FAILED! => changed=false 
  msg: 'BadRequestException: 400: Client Error for url: http://10.0.110.233:9696/v2.0/security-group-rules, Unrecognized attribute(s) ''remote_address_group_id'''

PR #351, opened 2021-01-25
Pin openstacksdk version due to os_security_group_rule module breakage

jistr:
The pin works around this breakage in Ansible:

'BadRequestException: 400: Client Error for url:
http://10.0.110.233:9696/v2.0/security-group-rules, Unrecognized
attribute(s) ''remote_address_group_id'''

Resolves: https://github.com/os-migrate/os-migrate/issues/350

### DIFF
--- a/toolbox/build/venv-requirements.txt
+++ b/toolbox/build/venv-requirements.txt
@@ -1,6 +1,7 @@
 # runtime
 ansible>=2.9.1,<2.10
-openstacksdk>=0.39.0
+# openstacksdk pin for https://github.com/os-migrate/os-migrate/issues/350
+openstacksdk>=0.39.0,<0.53
 
 # test
 ansible-lint==4.1.0


ccamacho:
Yeah I believed that my patch was breaking the CI but it's something generalized, if we move to 2.10 it will break. +1 to add the pin to unblock  the CI and then figure what changed.

ccamacho:
In theory, both parameters should be supported and working this is what I was testing last Friday (with the fqcn)

https://github.com/os-migrate/os-migrate/pull/352

jistr:
It looks like the remote_address_group_id is a newly added parameter in latest openstack, and the older APIs (including Train/16) don't understand it and treat it as an error when the parameter is provided. So this looks like a legit bug either in newest Ansible or newest openstacksdk. We're not explicitly setting the remote_address_group_id anywhere, but it still gets sent to API.

ccamacho:
If this passes, let's merge it.

ccamacho (approved):

PR #352, opened 2021-01-25
fix: remove remote_group_id in favor of remote_address_group_id

ccamacho:
Those 2 params should be compatible, let's use
the latest and allow having 2.10

### DIFF
--- a/os_migrate/plugins/module_utils/security_group_rule.py
+++ b/os_migrate/plugins/module_utils/security_group_rule.py
@@ -15,7 +15,7 @@ class SecurityGroupRule(resource.Resource):
     info_from_sdk = [
         'id',
         'security_group_id',
-        'remote_group_id',
+        'remote_address_group_id',
         'project_id',
         'created_at',
         'updated_at',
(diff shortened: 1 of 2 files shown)


jistr (changes requested):
I don't think one replaces the other, it's an addition. Going forward we should probably support both. https://specs.openstack.org/openstack/neutron-specs/specs/victoria/address-groups-support-in-security-group-rule.html

ccamacho:
I can't find the V docs for the openstacksdk but from latest:

But they look like the same https://docs.openstack.org/openstacksdk/latest/user/resources/network/v2/security_group_rule.html

```
remote_group_id
The remote security group ID to be associated with this security group rule. You can specify either remote_group_id or remote_address_group_id or remote_ip_prefix in the request body.

remote_address_group_id
The remote address group ID to be associated with this security group rule. You can specify either remote_group_id or remote_address_group_id or remote_ip_prefix in the request body.
```

jistr:
It's not the same though :). See the link i posted, security group is a group of firewall rules, but an address group is just a group of IP addresses. They are different resources on the Neutron API.

PR #355, opened 2021-01-27
Fix: Explicitly install nmap-ncat on RHEL

jistr:
It seems RPM dependency graphs have changed, and while we were getting
nmap-ncat installed via some dependency chain earlier, we need to
install it explicitly now.


ccamacho (approved):

jistr:
In another CI run i noticed that it was likely caused by a subscription error and not installing any packages, but that error was ignored. I think this PR might not be necessary. I posted #356 as a fix instead.

PR #356, opened 2021-01-27
Fix: Always stop on errors in conversion host deployment

jistr:
Previously when the source or destination conversion host encountered
some errror, Ansible would mark that host as failed and not run any
other tasks there, but it would continue to run tasks on other hosts,
most notably the migrator. This meant that e.g. subscription errors
wouldn't stop the playbook immediately, and if conversion host
deployment was included into a more complex migration playbook, the
migration itself would fail much later with a non-obvious error. This
is now fixed: whenever an error during conversion host deployment
happens, the playbook stops right away.

### DIFF
--- a/os_migrate/playbooks/deploy_conversion_hosts.yml
+++ b/os_migrate/playbooks/deploy_conversion_hosts.yml
@@ -1,4 +1,5 @@
 - hosts: migrator
+  any_errors_fatal: true
   tasks:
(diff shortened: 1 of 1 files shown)


ccamacho (approved):
lgtm

jistr:
I'll try again to see if we hit the same error or not.

PR #361, opened 2021-02-02
Dev: Lint fqcn

ccamacho:
This patch enforce the policy to use the FQCN
across the collection.

### DIFF
--- a/scripts/check_fqcn.py
+++ b/scripts/check_fqcn.py
@@ -1,0 +1,130 @@
+#!/usr/bin/env python
+
+"""
+Licensed under the Apache License, Version 2.0 (the "License"); you may
+not use this file except in compliance with the License. You may obtain
+a copy of the License at:
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+License for the specific language governing permissions and limitations
+under the License.
+"""
+
+import os
+import sys
+import urllib.request
+from pathlib import Path
+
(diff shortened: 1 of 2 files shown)


ccamacho:
After this pass Ill add a change to force the lint check to fail and remove that spurious file

ccamacho:
Failing as expected:

This is the debug message once there are non FQCN methods. 

There are some tasks using non FQCN methods: [(PosixPath('/root/os_migrate/scripts/../os_migrate/roles/export_projects/tasks/main.yml'), 'set_fact', 'no dot in namespace')]

scripts/check_fqcn.py
```diff
@@ -0,0 +1,130 @@
...
+                pass
+    return errors
+
+
+loader = DataLoader()
+inventory = InventoryManager(loader=loader, sources='localhost,')
+variable_manager = VariableManager(loader=loader, inventory=inventory)
+
+url = ("https://raw.githubusercontent.com/"
+       "ansible/ansible/devel/lib/ansible/config/"
```
ccamacho, line 111: This file do not exists in 2.9 that's why I'm pulling directly from it's content in github

jistr: Maybe we should pull the file from a tagged commit (2.10.x) rather than devel branch, In case it is moved or removed or changed unexpectedly? IIUC it's a part of an internal mechanism of Ansible, not user-facing one, so i'd assume it can change. Not a blocker though.

jistr (approved):

scripts/check_fqcn.py
odyssey4me: Yeah, I agree - better to pin it to a released/tagged version. This file will change quite a bit over time and will likely get removed for Ansible 3.x which is on the horizon.

ccamacho: done

PR #362, opened 2021-02-03
Chg: Preserve tags during migration of routers, networks, subnets

jistr:
Dev: Add after-update (and after-create) hook mechanism to resources

Since it is reasonable to expect that similar or the same actions will
take place in after create and after update, we have a single
`_hook_after_update` method which gets called in both events. It has
an `is_create` parameter in case distinguishing between the two events
is necessary.

------

Chg: Preserve tags during migration of routers, networks, subnets

The tags on networking resources cannot be set during the initial
creation request. Previously we weren't setting tags at all during
migration of networking resources. We are now migrating tags correctly
on routers, networks and subnets using a post-create/update hook.

### DIFF
--- a/tests/func/tenant/seed/router.yml
+++ b/tests/func/tenant/seed/router.yml
@@ -11,3 +11,20 @@
       - net: osm_net
         subnet: osm_router_subnet
         portip: 192.168.10.10
+  register: _osm_router
+
+- name: get auth token
+  openstack.cloud.auth:
+    auth: "{{ os_migrate_src_auth }}"
+
+- name: set tags on router
+  ansible.builtin.command:
+    cmd: openstack router set --tag test_tag_1 --tag test_tag_2 {{ _osm_router.id }}
+  environment:
+    OS_AUTH_TYPE: token
+    OS_AUTH_URL: "{{ os_migrate_src_auth.auth_url }}"
+    OS_TOKEN: "{{ auth_token }}"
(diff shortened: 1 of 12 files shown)


tests/func/tenant/seed/router.yml
```diff
@@ -11,3 +11,20 @@
...
+
+- name: set tags on router
+  shell: |
+    export OS_AUTH_TYPE=token
+    export OS_AUTH_URL="{{ os_migrate_src_auth.auth_url }}"
+    export OS_TOKEN="{{ auth_token }}"
+    export OS_PROJECT_ID="{{ os_migrate_src_auth.project_id|default('') }}"
+    export OS_PROJECT_NAME="{{ os_migrate_src_auth.project_name|default('') }}"
+    export OS_PROJECT_DOMAIN_ID="{{ os_migrate_src_auth.project_domain_id|default('') }}"
+    export OS_PROJECT_DOMAIN_NAME="{{ os_migrate_src_auth.project_domain_name|default('') }}"
```
odyssey4me, line 28: Are these exported using shell for debug value? If not, I'd suggest using the 'environment' argument on the task instead. Also, this could use command instead of shell.

jistr: Thanks, i'll try that.

jistr:
In addition to the discussed, i realized i'm not using FQCNs, will fix.

odyssey4me (approved):

Issue #343, opened 2021-01-21
AttributeError: 'Image' object has no attribute 'owner'


benyraja:
When trying to export the workloads, getting this error. I have imported the images ( verified the owner id is present ), networks, keypairs, flavors etc..

`The full traceback is:
Traceback (most recent call last):
  File "/root/.ansible/tmp/ansible-tmp-1611234052.7762973-811256-75477694867635/AnsiballZ_export_workload.py", line 102, in <module>
    _ansiballz_main()
  File "/root/.ansible/tmp/ansible-tmp-1611234052.7762973-811256-75477694867635/AnsiballZ_export_workload.py", line 94, in _ansiballz_main
    invoke_module(zipped_mod, temp_path, ANSIBALLZ_PARAMS)
  File "/root/.ansible/tmp/ansible-tmp-1611234052.7762973-811256-75477694867635/AnsiballZ_export_workload.py", line 40, in invoke_module
    runpy.run_module(mod_name='ansible_collections.os_migrate.os_migrate.plugins.modules.export_workload', init_globals=None, run_name='__main__', alter_sys=True)
  File "/usr/lib64/python3.6/runpy.py", line 205, in run_module
    return _run_module_code(code, init_globals, run_name, mod_spec)
  File "/usr/lib64/python3.6/runpy.py", line 96, in _run_module_code
    mod_name, mod_spec, pkg_name, script_name)
  File "/usr/lib64/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/tmp/ansible_os_migrate.os_migrate.export_workload_payload_bz6h3007/ansible_os_migrate.os_migrate.export_workload_payload.zip/ansible_collections/os_migrate/os_migrate/plugins/modules/export_workload.py", line 135, in <module>
  File "/tmp/ansible_os_migrate.os_migrate.export_workload_payload_bz6h3007/ansible_os_migrate.os_migrate.export_workload_payload.zip/ansible_collections/os_migrate/os_migrate/plugins/modules/export_workload.py", line 131, in main
  File "/tmp/ansible_os_migrate.os_migrate.export_workload_payload_bz6h3007/ansible_os_migrate.os_migrate.export_workload_payload.zip/ansible_collections/os_migrate/os_migrate/plugins/modules/export_workload.py", line 121, in run_module
[...]

PR #369, opened 2021-02-10
New: Testing with lower bound requirements

jistr:
We started testing with two OpenStack SDK versions:

* 0.52.x which is close to latest but not yet affected by a security
  group rules bug [1].

* 0.36.x which is in Train release (OSP 16).

[1] https://storyboard.openstack.org/#!/story/2008577

### DIFF
--- a/os_migrate/plugins/module_utils/image.py
+++ b/os_migrate/plugins/module_utils/image.py
@@ -131,7 +131,7 @@ class Image(resource.Resource):
 
     @staticmethod
     def _find_sdk_res(conn, name_or_id, filters=None):
-        # Glance requires owner_id instead of project_id.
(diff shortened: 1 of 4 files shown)


jistr:
The first e2e 10->16 is expected to fail -- we're now using the lower bound reqs file but the compatibility bug isn't fixed yet. Once we see the job fail here, i'll push another commit that should fix it.

jistr:
```
    AttributeError: 'Image' object has no attribute 'owner'
```

Yup the job failed as expected (as reported in #343 ). I'll push another commit to try and fix it.

jistr:
It works :)

odyssey4me (approved):

PR #371, opened 2021-02-12
Dev: Include the galaxy collection dependencies in the container.

ccamacho:
Installing the collection dependencies takes time, this PR
includes these dependencies in the os-migrate container image
so when they are installed if they are already include they will
be skipped.


jistr (approved):
In the commit message which disables Nova services CI we should explain why we're doing it. Not a blocker though.

jistr:
I will merge to unblock the CI for the other PR with sparsification fix.

PR #373, opened 2021-02-16
Dev: Enable nova services and functional tests

ccamacho:
This PR enables the functional tests related
to nova services.

### DIFF
--- a/.github/workflows/consistency-functional.yml
+++ b/.github/workflows/consistency-functional.yml
@@ -221,7 +221,7 @@ jobs:
           sudo chown root:root /bin
       - name: Clone devstack
         run: |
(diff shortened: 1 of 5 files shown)


ccamacho:
I think the root cause of the issue is that some config parameters are/were not working properly in latest/master. Ill push another PR moving this to master if we agreed to (otherwise we will need to fix it later when victoria is EOL).

jistr (approved):
Again, i think next time we'd better mention more rationale in the commit message - "master looks broken, let's pin to victoria branch". The fact that we're pinning to Victoria is only visible if you look at the commit diff, but such a significant change should be visible from the commit message. Not blocking though.

PR #381, opened 2021-03-02
Dev: Run shell when venv-run is passed no arguments

jistr:
In toolbox dir, we have the run/shell script pairs, one for running a
command in a container, another for launching a shell in a
container. However, it can be done so that if no arguments are passed
to the 'run' script, it acts the same way as the 'shell' script. This
is already true for the plain 'run' and the 'vagrant-run', only
'venv-run' is edited to achieve the behavior described above.

### DIFF
--- a/toolbox/build/venv-wrapper
+++ b/toolbox/build/venv-wrapper
@@ -4,4 +4,8 @@ set -euo pipefail
 
 source /root/venv/bin/activate
 
-"$@"
+if [ "$#" = "0" ]; then
+    bash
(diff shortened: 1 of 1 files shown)


ccamacho (approved):

jistr:
E2E jobs passed, but the GH actions looked stalled and i didn't find a way to retrigger, so i just ran git commit --amend and re-pushed. If they pass, i think we don't need to re-run e2e tests here.

PR #384, opened 2021-03-10
New: Validate resource data structure

jistr:
When resources are being loaded, we now validate that the resource has
all the expected `params` and `_migration_params` and `_info.id`.
Errors are reported for resources that cannot be parsed. The resources
in error messages are identified by type+name+id tuple, if those
attributes of the resource aren't missing.


ccamacho (approved):
LGTM

PR #392, opened 2021-04-01
New: Ensure Nova server creation is successful and emit detailed info

jistr:
The import_workload_create_instance module now outputs a full workload
serialization of the server created in the destination cloud. When
running Ansible in verbose mode, it can be used as debugging
information, and allows for direct comparison of the destination
workload to the contents of workloads.yml file.

This requires that the import_workload_create_instance module waits
for the Nova server to enter ACTIVE state, to have full information
available. Typically this only adds several seconds to the workload
migration time, which is negligible in the full picture. By waiting,
we can also make sure the server indeed does go ACTIVE and not into
ERROR state, which may have previously gone unnoticed with the "fire
and forget" server creation approach.

### DIFF
--- a/os_migrate/plugins/module_utils/server.py
+++ b/os_migrate/plugins/module_utils/server.py
@@ -179,7 +179,7 @@ class Server(resource.Resource):
 
         # There are multiple representations of server in SDK, some of
         # them don't have flavor ID info.
-        flavor_id = sdk_res['flavor'].get('id')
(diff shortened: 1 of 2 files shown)


jistr:
Now the CI errors are different - package download failure, github API failure. Trying again...

jistr:
Every time a different error, will keep retrying until we see a pattern.

ccamacho (approved):

PR #394, opened 2021-04-19
New: Configurable timeout for non-immediate actions

jistr:
Until now, non-immediate actions and blocking calls to OpenStack SDK
within workload migration have used a hard-coded timeout of 30
minutes. New variable `os_migrate_timeout` has been introduced for
setting the timeout (in seconds), the value is 1800 (= 30 minutes).


ccamacho (approved):

jistr:
Tested the negative scenario by setting the timeout very low, it works:

```
    openstack.exceptions.ResourceTimeout: Timeout waiting for the volume to be available.
```

PR #410, opened 2021-05-18
fix: increase timeouts when deployment takes more than default

ccamacho:
This commit allows to increase the hardwired default timeout of
600 seconds when deploying the conversion hosts.

In this case, when the overlay network provider takes longer to
assign a floating IP, there might be needed to increase this
default.


jistr (approved):

jistr:
I just discovered ansible-test tests were ignored and broken, let's please first merge https://github.com/os-migrate/os-migrate/pull/412 and then rebase this PR, so that we ensure this PR doesn't introduce new test/lint problems.

jistr (approved):

ccamacho:
Failed with

```
keystoneauth1.exceptions.http.InternalServerError: Internal Server Error (HTTP 500)
```

PR #411, opened 2021-05-18
New: delete volumes when deleting the conversion hosts

ccamacho:
This commit adds the hability to remove automatically
the volume created with the conversion host once
the VM is destroyed.


jistr:
I just discovered ansible-test tests were ignored and broken, let's please first merge https://github.com/os-migrate/os-migrate/pull/412 and then rebase this PR, so that we ensure this PR doesn't introduce new test/lint problems.

jistr (approved):

PR #412, opened 2021-05-21
Dev: Fix running tests via Ansible test tools

jistr:
All tests running via the `ansible-test` tool were not run for some
time. This was due to Ansible looking for specific Python versions to
run the tests with. The latest version it was looking for was Python
3.8, but in the toolbox container we had Python 3.9. This caused
ansible-test to silently skip running all tests and still return 0, so
this went completely unnoticed in CI.

The toolbox container now explicitly install Python 3.8 and uses it to
create the virtualenv. This caused all the tests to start running
again, and they failed, so minimal fixes to make them pass again are
included in the commit too.


ccamacho (approved):
The 13->16 passed we should just wait for the job to write the result.

PR #413, opened 2021-05-21
Fix: Export RouterInterface of HA and DVR routers

jistr:
Previously when exporting router interfaces, only ports with
device_owner value of 'network:router_interface' were picked up. This
would not export interfaces correctly for HA/DVR routers. We now also
export ports with device_owner values of
'network:ha_router_replicated_interface' and
'network:router_interface_distributed' for HA/DVR routers.


ccamacho (approved):

jistr:
Tested also manually to the best extent with the clouds i have access to, seems to work ok, merging.

PR #414, opened 2021-05-25
Dev: Mount resolv.conf into toolbox

jistr:
This makes it possible to resolve DNS names in a VPN for example.


ccamacho (approved):

jistr:
This had e2e green and there were no changes to the patch, i only rebased onto the fixed func CI. Merging.

PR #417, opened 2021-05-25
dev: fix CI, disable neutron agents not used with OVN

ccamacho:
DEvstack now uses OVN, this commit disables those
agents not needed.


jistr:
Awesome!

jistr (approved):

PR #419, opened 2021-05-27
fix: Do not check for the subscribed_pool_ids 

ccamacho:


ccamacho:
The 10->16 job uses RHEL as the conversion host.

ccamacho:
Our CI has no subscribed_pool_ids because we are using an internal instance. We will check then for `System successfully registered`:

```
TASK [os_migrate.os_migrate.conversion_host_content : log the subscription manager output] ***
Thursday 27 May 2021  17:12:46 +0000 (0:00:46.172)       0:04:28.592 ********** 
ok: [os_migrate_conv_src] => 
  msg:
    changed: true
    failed: false
    msg: System successfully registered to 'subscription.rhn.stage.redhat.com'.
    subscribed_pool_ids: []
```

jistr (approved):
Looks good and the CI went past the point it seems. +1

ccamacho:
I'm seeing some weird behavior with this, please let me confirm is good to be merge before merging.

PR #423, opened 2021-06-01
Dont run everywhere

ccamacho:


philroche (approved):
This all looks good. 

Not too familiar with github action workflows but based on 
 https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context  and https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idif I am +1

jistr:
Looks good but both need changed commit messages from "Chg" to "Dev", we don't want this kind of info bubbling up into changelog, which should be user-facing.

When selecting the commit message between Dev and the others (New/Chg/Fix/Oth), the question to ask is: "Does this commit change something about usage of OS Migrate in production scenarios?". If the answer is no, the commit always has to be "Dev".

jistr:
Adding extra detail as we have @philroche onboarding: we have an example where we let a commit message slip recently in the unreleased changelog: https://os-migrate.github.io/os-migrate/changelog.html

"Lint tests are failing with the line lenght <=100. This commits increases the limit to 150."  -- not really something user wants to read when they are curious what changed in the new release since the last one :)

jistr:
The changelog generation is explained in dev docs https://os-migrate.github.io/os-migrate/devel/contributing-general.html#commit-messages-for-changelog

jistr (approved):

PR #425, opened 2021-06-03
Dev: skip file-changes-action on the scheduled job

ccamacho:
This commit skips the execution of the file-changes-action
GitHub action when running the scheduled job. This is because
the file-changes-action action waits for a PR id to determine
which files changed, and when running the scheduled job this
variable will be NaN.


ccamacho:
This also fixes a new breaking change in devstack

ccamacho:
Passed, I'm just amending one title task to make it self explanatory as we miss the task name

jistr (approved):
We probably can't test the scheduled job right here? But the changes look good. Thanks for fixing the CI :).

PR #428, opened 2021-06-04
Fix: Flavor extra_specs import with OpenStack SDK older than 0.51

jistr:
The previous implementation of extra_specs import on flavors only
worked with SDK >=0.51. A fallback implementation is now added for
running with older SDKs. This fallback implementation can be dropped
in the future after we drop support for SDK <0.51.

### DIFF
--- a/os_migrate/plugins/module_utils/flavor.py
+++ b/os_migrate/plugins/module_utils/flavor.py
@@ -45,11 +45,21 @@ class Flavor(resource.Resource):
         delete_extra_specs = list(set(sdk_extra_specs.keys()) -
                                   set(params_extra_specs.keys()))
         for prop_name in delete_extra_specs:
-            conn.compute.delete_flavor_extra_specs_property(sdk_res, prop_name)
+            # We can use the 'if' variant exclusively after we mandate OpenStack SDK >= 0.51
+            if hasattr(conn.compute, 'delete_flavor_extra_specs_property'):
+                conn.compute.delete_flavor_extra_specs_property(sdk_res['id'], prop_name)
+            else:
+                # Old non-proxy SDK API
+                conn.unset_flavor_specs(sdk_res, [prop_name])
         for prop_name in params_extra_specs.keys():
             if sdk_extra_specs.get(prop_name) != params_extra_specs[prop_name]:
-                conn.compute.update_flavor_extra_specs_property(
-                    sdk_res, prop_name, params_extra_specs[prop_name])
+                # We can use the 'if' variant exclusively after we mandate OpenStack SDK >= 0.51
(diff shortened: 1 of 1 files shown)


philroche (approved):
Changes look good as per docs. 

Non blocking comments: It looks like you could omit the first for loop and call `conn.unset_flavor_specs(sdk_res, delete_extra_specs)` which would work for both > 0.51 and < 0.51. 

See https://github.com/openstack/openstacksdk/blob/0.50.0/openstack/cloud/_compute.py#L1475 

The same goes for calling `set_flavor_specs`.

Or is the goal to use `delete_flavor_extra_specs_property` and `update_flavor_extra_specs_property` if they are available even if they are not being used in `unset_flavor_specs` and `set_flavor_specs` yet like in 0.51? From what I can see `unset_flavor_specs` and `set_flavor_specs` were updated to use `delete_flavor_extra_specs_property` and `update_flavor_extra_specs_property` in 0.5.3.

jistr:
Yea i think one of the goals would be to use the "modern" APIs, so nothing which is directly on the `conn` object if possible. My understanding is the APIs on the "service proxies" get new features while the APIs on the `conn` object only get minimal maintenance to not break. E.g. we could only add support for `description` of Nova Servers after we switched from `conn.create_server` to `conn.compute.create_server`:

https://github.com/os-migrate/os-migrate/commit/fd608b98d4fb9a91cd018ae668cd0a97aac25a94#diff-6466a6706ae1feebb369885703bb83b331ec31b4dde467e2d3477924d97343be

So i tried to structure the code in a way that later when we remove the `conn.(un)set_flavor_specs` functions, we don't have to change the control flow code.

jistr:
Meh wrong link above, i meant to simply link the full commit: https://github.com/os-migrate/os-migrate/commit/fd608b98d4fb9a91cd018ae668cd0a97aac25a94

philroche:
> Yea i think one of the goals would be to use the "modern" APIs, so nothing which is directly on the `conn` object if possible. My understanding is the APIs on the "service proxies" get new features while the APIs on the `conn` object only get minimal maintenance to not break. E.g. we could only add support for `description` of Nova Servers after we switched from `conn.create_server` to `conn.compute.create_server`:
> 
> [fd608b9#diff-6466a6706ae1feebb369885703bb83b331ec31b4dde467e2d3477924d97343be](https://github.com/os-migrate/os-migrate/commit/fd608b98d4fb9a91cd018ae668cd0a97aac25a94#diff-6466a6706ae1feebb369885703bb83b331ec31b4dde467e2d3477924d97343be)
> 
> So i tried to structure the code in a way that later when we remove the `conn.(un)set_flavor_specs` functions, we don't have to change the control flow code.

Good context. Thank you

PR #429, opened 2021-06-07
Fix: Only use 'type' on keypairs when it is supported

jistr:
SDK should support 'type' on keypairs since 0.32 [1], however even
with 0.36 this error still appears when testing against latest
Devstack:

openstack.exceptions.BadRequestException: BadRequestException: 400:
Client Error for url: http://192.168.122.93/compute/v2.1/os-keypairs,
Invalid input for field/attribute keypair. Value: {'type': 'ssh',
<REDACTED>, 'name': 'osm_keypair'}. Additional properties are not
allowed ('type' was unexpected).

This error disappears when testing with latest SDK. It's not easy to
track down what exact combinations of SDK/API are needed to allow for
'type' in keypair creation request. So we'll first try with 'type',
and if that request fails with error 400 (Bad Request), then we retry
without 'type'. If that request still fails, the keypair import module
fails too.

[1] https://github.com/openstack/openstacksdk/commit/15baef656ac56421a71e691982a70b218110f18d

### DIFF
--- a/os_migrate/plugins/module_utils/keypair.py
+++ b/os_migrate/plugins/module_utils/keypair.py
@@ -29,7 +29,11 @@ class Keypair(resource.Resource):
 
     @staticmethod
     def _create_sdk_res(conn, sdk_params):
-        return conn.compute.create_keypair(**sdk_params)
+        try:
+            return conn.compute.create_keypair(**sdk_params)
+        except openstack.exceptions.BadRequestException:
+            sdk_params_no_type = {k: v for k, v in sdk_params.items() if k != 'type'}
+            return conn.compute.create_keypair(**sdk_params_no_type)
 
     @staticmethod
     def _find_sdk_res(conn, name_or_id, filters=None):


jistr:
I'd rather do a check based on client (or API?) version but there doesn't appear to be a well defensible rationale of "i will only include 'type' in the request when <some condition>", so i opted for "try it and fall back if needed" approach.

os_migrate/plugins/module_utils/keypair.py
```diff
@@ -29,7 +29,11 @@ class Keypair(resource.Resource):
 
     @staticmethod
     def _create_sdk_res(conn, sdk_params):
-        return conn.compute.create_keypair(**sdk_params)
+        try:
+            return conn.compute.create_keypair(**sdk_params)
+        except openstack.exceptions.BadRequestException:
+            sdk_params_no_type = {k: v for k, v in sdk_params.items() if k != 'type'}
```
philroche, line 35: Would using `del` be a bit cleaner?

```
if 'type' in sdk_params.keys():
    # With BadRequestException it is likley that 'type' is not yet supported. Remove and try again
    del sdk_params['type']
```

philroche (commented):
+1 on reasoning. One small query on the implementation?

os_migrate/plugins/module_utils/keypair.py
philroche: Thank you for the reasoning. I couldn't immediately see any other use of `sdk_params` but that doesn't mean it won't happen in the future. +1 on your change.

philroche (approved):

jistr:
Thanks, we won't have Carlos for a while so i'll go ahead and merge.

PR #430, opened 2021-06-09
Fix: Remove unnecessary delay in conversion host reachability check

jistr:
In conversion host reachability check, with the `wait_for` Ansible
module we used `delay` of 10 seconds, which caused a delay before the
first check, causing unnecessary slowdown (20 seconds in total) of the
import_workloads playbook. What we probably meant to use was `sleep`,
which sets the interval between checks. This is now fixed and the
`sleep` property is set to 5 seconds.


philroche (approved):
+1 and as per docs https://docs.ansible.com/ansible/latest/collections/ansible/builtin/wait_for_module.html

jistr:
Thanks :)

PR #431, opened 2021-06-10
Prepare release 0.8.3 with recent fixes

jistr:
Fix: Handle when flavor extra_specs are reported as None from API
Fix: Flavor extra_specs import with OpenStack SDK older than 0.51
Fix: Only use 'type' on keypairs when it is supported
Fix: Remove unnecessary delay in conversion host reachability check

+ 'Dev' patches which look like they may be required to keep tests passing


philroche (approved):
+1

jistr:
Thanks :)

PR #432, opened 2021-06-15
Dev: Expand prerequisites for developer environment setup

philroche:
Add links to Podman and Buildah for easier discovery and also include
the prerequisite that a git clone of the os-migrate repo is also required

### DIFF
--- a/docs/src/devel/dev-env-setup.rst
+++ b/docs/src/devel/dev-env-setup.rst
@@ -4,7 +4,8 @@ Development Environment Setup
 Prerequisites
 -------------
 
--  Podman and Buildah for running rootless development environment
+-  Local clone of the os-migrate git repo from https://github.com/os-migrate/os-migrate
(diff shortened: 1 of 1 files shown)


docs/src/devel/dev-env-setup.rst
```diff
@@ -4,7 +4,8 @@ Development Environment Setup
 Prerequisites
 -------------
 
--  Podman and Buildah for running rootless development environment
+-  Local clone of the os-migrate git repo from https://github.com/os-migrate/os-migrate
+-  `Podman <https://podman.io/>`_ and `Buildah <https://buildah.io/>`_ for running rootless development environment
```
jistr, line 8: Ideally we should try to keep some line wrap line (72/80 chars or something like that). Not a problem though. Merging, thanks :).

jistr (approved):

docs/src/devel/dev-env-setup.rst
philroche: Noted. Thanks

PR #435, opened 2021-06-16
Dev: Development setup updates

philroche:
* Dev: Typo fix in dev env setup docs - resoruces -> resources

* Dev: Update docs for development environment credential creation

Development environment functional tests now require two credential
file: `tests/func/auth_tenant.yml` and `tests/func/auth_admin.yml`

These are created by running the

`./toolbox/run make test-setup-vagrant-devstack` and
`./toolbox/run make test-setup-vagrant-devstack-admin` respectively.

* Dev: Maintain consistent line wrap in docs

Last PR for docs updates
https://github.com/os-migrate/os-migrate/pull/432
exceeded the recommended line wrap. Fixing now

* Dev: Set a consistent default for the NO_VAGRANT variable

If vagrant is not required for testing one can set `NO_VAGRANT=1`.
The default for NO_VAGRANT was either 0 or ''. This commit sets
NO_VAGRANT's default as 0 consistently.

This also removes the

`[Warning] one or more build args were not consumed: [NO_VAGRANT]`

warning when building the toolbox container.

* Dev: set default for XDG_MENU_PREFIX environment variable

XDG_MENU_PREFIX is used in vagrant setup during developer environment setup

If not set then setup will fail with

```
./toolbox/vagrant-run: line 49: XDG_MENU_PREFIX: unbound variable
```

Some desktop environments, like Mate on Fedora 33, do not set this variable
so we should set a default to avoid this error.


jistr (approved):
Looks good, thanks!

jistr:
I just realized, since we're changing Dockerfile and build.sh, we should run the e2e test suite too. The change looks pretty safe though, especially given that the other jobs passed.

PR #436, opened 2021-06-24
New: Export+import of user Nova keypairs as admin 

jistr:
(detail in commit messages)

### DIFF
--- a/os_migrate/plugins/module_utils/reference.py
+++ b/os_migrate/plugins/module_utils/reference.py
@@ -322,6 +322,55 @@ def project_ref(conn, id_, required=True):
     return ref
 
 
+def user_id(conn, ref, required=True, none_if_auth=False):
+    """Fetch ID of a User identified by reference dict `ref`. Use
+    OpenStack SDK connection `conn` to fetch the info. If `required`,
+    ensure the fetch is successful. If `none_if_auth` and the user
+    name is '%auth%', return None instead of the actual user ID
+    (useful for skipping the parameter).
+    Returns: the ID, or None if not found and not `required`
+    Raises: openstack's ResourceNotFound when `required` but not found
+    """
+    if ref['name'] == '%auth%':
+        if none_if_auth:
+            return None
+        else:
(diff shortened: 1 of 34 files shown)


os_migrate/playbooks/export_user_keypairs.yml
```diff
@@ -0,0 +1,3 @@
+- hosts: migrator
+  roles:
+    - os_migrate.os_migrate.export_user_keypairs
```
jistr, line 3: I think i'll still rename the playbook and role to `export_users_keypairs`, meaning "export users' keypairs". I'd like it to be more obvious from the playbook name that when executed, it will do an `O(n^2)` type of operation.

os_migrate/plugins/module_utils/reference.py
```diff
@@ -322,6 +322,55 @@ def project_ref(conn, id_, required=True):
...
+    OpenStack SDK connection `conn` to fetch the info. If `required`,
+    ensure the fetch is successful. If `allow_auth`, return '%auth%'
+    reference in case the looked up user ID is the same as currently
+    authenticated user ID.
+    Returns: the ref dict, or None if not found and not `required`
+    Raises: openstack's ResourceNotFound when `required` but not found
+    """
+    if allow_auth and id_ == conn.current_user_id:
+        return {
+            'project_name': '%auth%',
```
jistr, line 354: In a separate PR it would be worth moving this '%auth%' magic value into consts.py. It is already used in different places across OS Migrate, so i won't piggy-back the change into this commit.

jistr:
Export only for now, will work on import.

jistr:
All green but still not done. Just by chance i ran across an issue where having same-named keys in different problems causes issues. OS Migrate is wired for the name-based identity checks so i need to fix the identity mechanism for keypairs and the general pre-import validation.

philroche (approved):
Well documented and straight forward to follow.

jistr:
Thanks for the review, merging.

PR #438, opened 2021-07-05
Dev: Force disable ironic

ccamacho:
This commit disables ironic.


ccamacho:
@jistr @philroche the Devstack deployment is failing in the periodic check, we need to force disable ironic this time ;)

ccamacho:
nope, looks like is passing now :) sy!

PR #440, opened 2021-07-07
Dev: Add minimal runtime.yml for the Ansible collection

jistr:
Publishing 0.9.0 failed with:

ERROR! Galaxy import process failed: 'requires_ansible' in
meta/runtime.yml is mandatory, but no meta/runtime.yml found (Code:
None)

It seems there is a new requirement for Ansible collections of adding
meta/runtime.yml [1], and the minimal runtime.yml seems to consist of
supported range of Ansible versions. Let's start defensively with 2.9
only, as that's what we currently CI, and expand/move the list
according to what we test.

[1] https://github.com/ansible/ansible/commit/a9eb8b04882669bd17cd82780c908863e7504886

### DIFF
--- a/os_migrate/meta/runtime.yml
+++ b/os_migrate/meta/runtime.yml
@@ -1,0 +1,1 @@
+requires_ansible: ">=2.9,<2.10"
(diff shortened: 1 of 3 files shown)


philroche (approved):
LGTM

jistr:
Thanks :)

PR #443, opened 2021-07-19
Fix: Glance API attribute 'stores' is read-only now

philroche:
Due to bug https://bugs.launchpad.net/glance/+bug/1889676 glance now
sets stores as a read only property. As such we should remove it from
any create image calls.

See https://github.com/openstack/glance/commit/e1f0e94b90b162b1cdca5bdd7af3c09a90ca80ac#diff-3a53e94e6a14f08e4e3305f4e4388819cef4f35a9277afeab2f07bc2664489d6
for the Glance API change and also see conversation on
terraform-provider-openstack repo @
https://github.com/terraform-provider-openstack/terraform-provider-openstack/issues/1091#issuecomment-806003499
for some related context.

This bug was found when testing e2e tests using an infrared deployed Openstack 16.1.


os_migrate/plugins/module_utils/image.py
```diff
@@ -89,6 +89,14 @@ class Image(resource.Resource):
         'ramdisk_id',
     ]
 
+    # Some parameters are allowed when reading an image resource but not when
+    # creating or updating it.  This list of parameter names is purged from the parameter
+    # list before calling create or update and removed from object properties when reading from sdk
+    # Attribute 'stores' is read-only now. Due to bug https://bugs.launchpad.net/glance/+bug/1889676 glance now
+    # sets stores as a read only property. As such we should remove it from any create or update calls.
+    # An existing pattern is to use the readonly_sdk_params list to do this.
+    readonly_sdk_params = ["store", "stores"]
```
jistr, line 98: The readonly_sdk_params are related to what is directly in the 'params' dictionary (if you look at the serialization in images.yml, or how readonly_sdk_params is used in resource.py). Since we've removed 'store' from 'params' into '_info' above, we no longer need this part.

Re using this to filter 'properties' below on line 183 - i wouldn't double this variable's meaning and use it also as a filter for the 'properties' key in 'params' dictionary. IMO we should have one variable for filtering `['params']` (we do have it inherited from resource.py but no longer need to override it here) and another variable for filtering `['params']['properties']`. I do think it's a good idea to have a variable for this, could be class-wide but maybe having the variable locally in a function that will do the filtering would suffice?

os_migrate/plugins/module_utils/image.py
```diff
@@ -89,6 +89,14 @@ class Image(resource.Resource):
...
+
     @classmethod
     def from_sdk(cls, conn, sdk_resource):
         obj = super().from_sdk(conn, sdk_resource)
```
jistr, line 102: We do filtering out of ['properties']['self'] in this function already. This means it never gets serialized out into 'params' in the first place. This should work better than filtering on import, because whatever is in 'params' plays a role in idempotency checking. So if we're not using 'self' or 'store' when creating the resource, it should not be in 'params' at all, as it could be causing "false alarms" when checking whether a resource needs updating in destination cloud or not.

So maybe we can put the variable here and filter out 'self', 'store', 'stores'? (I'm not sure about the 'store' vs. 'stores' difference and if we need both?)

jistr (changes requested):

os_migrate/plugins/module_utils/image.py
jistr: Sorry i got things mixed up here. On a second look, the way you used the variable does not mean we're doubling its meaning, as it's generally applied to the final `sdk_params` rather than the resource's `params`, so this was fine.

Still, i think we should probably do the filtering on export as mentioned below due to how idempotency is handled, and in that case we shouldn't use this variable as we won't be filtering the sdk_params anymore.

philroche:
@jistr All good points. Moving back to draft now

os_migrate/plugins/module_utils/image.py
```diff
@@ -93,10 +93,25 @@ class Image(resource.Resource):
...
+        #
+        # * 'stores'
+        #   Attribute 'stores' is read-only now. Due to bug https://bugs.launchpad.net/glance/+bug/1889676 glance now
+        #   sets stores as a read only property. As such we should remove it from any create or update calls.
+        # * 'self'
+        #   We need to remove 'self' from properties as it would break
+        #   idempotency check and it's not really a property anyway.
+        readonly_properties = ['self', 'stores']
+
+        # cast .keys() to list to avoid 'dictionary changed size during iteration' error
```
jistr, line 111: Nice

jistr (approved):
Thanks, looks good, let's wait for CI and then we can merge.

PR #447, opened 2021-07-26
Dev: Add developer documentation for os-migrate e2e tests

philroche:
Adding new developer documentation for the end to end tests
detailing the prerequisites, config required, expected output
and optional tags/variables to pass when running.

### DIFF
--- a/docs/src/devel/dev-env-setup.rst
+++ b/docs/src/devel/dev-env-setup.rst
@@ -178,3 +178,216 @@ e.g.:
 ::
 
    OS_MIGRATE_FUNC_TEST_ARGS='--tags test_network,test_subnet --skip-tags test_clean_after' ./toolbox/run make test-func
+
+Running e2e tests
+-----------------
+
+OS Migrate also has a suite of end to end, e2e, tests which tests a migration from one existing openstack deployment
+to another existing openstack deployment.
+
+You can also test with a single existing openstack deployment migrating from one project to another.
+
+These docs cover the testing of a single existing openstack deployment migrating from one project to another scenario.
+
+The concepts and prerequisites are the same for other deployments.
+
+Prerequisites for e2e test
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* Source environment credentials
+* Destination environment credentials
+* Existing images in both source and destination environments
+* Flavors
+* Space requirements
+  - 2 images totalling 1.25GB
+  - 1 volume totalling 1GiB in source environment
+  - 2 volumes totalling 5GiB in destination environment
+  - 2 VMs totalling 35GB disk usage in each environment
+
+Below are the steps required to satisfy the above requirements and run e2e tests in a test environment, migrating
+resources from one project to another.
+
+Create source environment and destination environment projects and users
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. code-block::
+
+    # Create the src user in the default domain with password 'redhat'
+    openstack user create --domain default --password redhat src
+
+    # Create the src project
+    openstack project create --domain default src
+
+    # Assign src user a 'member' role in the src project
+    openstack role add \
+    --user src --user-domain default \
+    --project src --project-domain default member
+
+    # Confirm role assignment was successful
+    openstack role assignment list --project src
+
+    # Create the dst user in the default domain with password 'redhat'
+    openstack user create --domain default --password redhat dst
+
+    # Create the dst project
+    openstack project create --domain default dst
+
+    # Assign dst user a 'member' role in the src project
+    openstack role add \
+    --user dst --user-domain default \
+    --project dst --project-domain default member
+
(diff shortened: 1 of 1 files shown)


docs/src/devel/dev-env-setup.rst
```diff
@@ -178,3 +178,216 @@ e.g.:
...
+    # Confirm role assignment was successful
+    openstack role assignment list --project dst
+
+Create images
+------------
+
+.. code-block::
+
+    # Download images
+    wget https://cloud.centos.org/centos/8/x86_64/images/CentOS-8-GenericCloud-8.2.2004-20200611.2.x86_64.qcow2
```
ccamacho, line 249: Can we use stream instead? We are currently testing it in the CI but I already deploy it.

docs/src/devel/dev-env-setup.rst
```diff
@@ -178,3 +178,216 @@ e.g.:
...
+to another existing openstack deployment.
+
+You can also test with a single existing openstack deployment migrating from one project to another.
+
+These docs cover the testing of a single existing openstack deployment migrating from one project to another scenario.
+
+The concepts and prerequisites are the same for other deployments.
+
+Prerequisites for e2e test
+--------------------------
```
ccamacho, line 195: shouldn't this be a subsection of Running e2e tests ?

ccamacho: same with the next subsections

docs/src/devel/dev-env-setup.rst
philroche: Good idea. Now updated

docs/src/devel/dev-env-setup.rst
philroche: Subsection now more obvious using '^' as per other docs

docs/src/devel/dev-env-setup.rst
```diff
@@ -178,3 +178,216 @@ e.g.:
...
+    os_migrate.os_migrate.import_workloads : clean up in the source cloud after migration ---------------------- 21.16s
+    os_migrate.os_migrate.conversion_host : create os_migrate conversion host ---------------------------------- 20.03s
+    Remove osm_server ------------------------------------------------------------------------------------------ 19.01s
+    os_migrate.os_migrate.conversion_host : create os_migrate conversion host ---------------------------------- 18.23s
+    Shutdown osm_server ---------------------------------------------------------------------------------------- 18.14s
+    Shutdown osm_server ---------------------------------------------------------------------------------------- 17.88s
+
+Optional tags to pass to e2e tests
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+There are a set of tags that can be used to filter which playbooks to run during test.
```
jistr, line 361: I'd say "which playbooks" -> "which plays" or "which tasks". I'd go with the latter as it's more likely to be correct in any circumstances.

docs/src/devel/dev-env-setup.rst
```diff
@@ -178,3 +178,216 @@ e.g.:
...
+- test_image_workload_boot_copy
+- test_image_workload_boot_nocopy
+- test_image_workload_boot_copy_clean
+- test_clean_before
+- test_pre_workload
+
+Optional playbook variable
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+There is also an optional variable `test_clean_conversion_hosts_after` which can be set to `false` if you do not wish
+to cleanup after test is complete.
```
jistr, line 374: "to clean up" -> "to clean up conversion hosts"

(Other cleanup aspects are controlled via the tags above. The conversion hosts cannot be controlled by tags because they're deleted via importing the user-facing playbook which does not have any `test_` tags, and the import_playbook task itself cannot be tagged, so we skip it based on a variable.)

jistr (commented):
Commented a few minor things that would be nice to tweak before merging, but these are great docs, thanks!

docs/src/devel/dev-env-setup.rst
philroche: Good clarification. Updated

docs/src/devel/dev-env-setup.rst
philroche: Thank you

docs/src/devel/dev-env-setup.rst
```diff
@@ -178,3 +178,216 @@ e.g.:
 ::
 
    OS_MIGRATE_FUNC_TEST_ARGS='--tags test_network,test_subnet --skip-tags test_clean_after' ./toolbox/run make test-func
+
+Running e2e tests
+-----------------
+
+os-migrate also has a suite of end to end, e2e, tests which tests a migration from one existing openstack deployment
```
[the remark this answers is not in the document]
philroche: Good point. Updated now

jistr:
Thanks. Can we squash the docs commits into a single one as we talked last time re the conventions we're presently using, so that we don't seep review micro-details into git logs? (Can be done quite easily with `git rebase --interactive`.)

philroche:
Commits now squashed as requested

jistr (approved):
Thanks :)

PR #450, opened 2021-07-27
New: Customizable attached volume properties in workload migration

jistr:
Attached volumes are now exported in the 'params' section of workload
serialization, as nested resources. This allows editing these desired
properties of attached volumes that should be created in the
destination cloud: name, description, availability zone, volume type.


tests/e2e/tenant/run_workload.yml
```diff
@@ -16,6 +16,8 @@
         os_migrate_workloads_filter:
           - regex: '^osm_(?!uch)'
 
+    - ansible.builtin.pause:
```
jistr, line 19: Nope

jistr:
This is how the new `volumes` parameter with nested resources looks like:

```yaml
    volumes:
    - _info:
        attachments:
        - attached_at: '2021-07-22T12:14:23.000000'
          attachment_id: e85cf712-7a44-4877-980a-732d396ece1d
          device: /dev/vdb
          host_name: compute-1.jistr-13.local
          id: 7e9373b3-6518-4722-ba6f-93e03945e766
          server_id: 5a8bc82d-2c34-4d02-9043-8eb3c99bc66b
          volume_id: 7e9373b3-6518-4722-ba6f-93e03945e766
        id: 7e9373b3-6518-4722-ba6f-93e03945e766
        is_bootable: false
        size: 1
      _migration_params: {}
      params:
        availability_zone: nova
        description: null
        name: osm_volume
        volume_type: tripleo
      type: openstack.network.ServerVolume
```

ccamacho (approved):

jistr:
@njnatale-lm Yes that's correct, such a change should be possible using this mechanism. You'd be looking at the `volumes` parameter if you wanted to change AZ for volumes which already existed with the source workload (any attached volumes, and boot volume in case the source server is boot-from-volume).

In case the source server is boot-from-image and you want to copy its boot disk contents (`boot_disk_copy: true`), then no volume exists in the source cloud (and thus isn't mentioned in `volumes` parameter) but a volume will be created in the destination cloud. It would use Cinder's default AZ by default, but you could set the AZ for this scenario explicitly at `_migration_params.boot_volume_params`. This part was implemented in #456 . Docs for it are at https://os-migrate.github.io/os-migrate/user/migration-params-guide.html#workload-migration-parameters .

PR #451, opened 2021-07-27
New: Allow ssh access to conversion host using a default password

ccamacho:
This commit allows to configure a default password in both
centos and cloud-user users, this can help connecting directly
to the hosts with a default password disabled by default.

Added os_migrate_conversion_host_ssh_user_enable_password_access
disabled by default.

### DIFF
--- a/os_migrate/playbooks/deploy_conversion_hosts.yml
+++ b/os_migrate/playbooks/deploy_conversion_hosts.yml
@@ -70,6 +70,18 @@
 - hosts: conversion_hosts
   any_errors_fatal: true
   tasks:
+    - name: enable ssh_user password access to the conversion hosts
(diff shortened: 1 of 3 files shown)


ccamacho:
The rationale for this addition is that if there is a failure deploying the conversion hosts, the keys used to access and provisioning them will be lost (in the case of our CI or anyone deploying from a container). For such cases where we are interested in debugging the conversion hosts, it will be possible now to enable password access to these machines and logging in using a pre-defined password.

jistr (approved):
LTGM, this will be helpful for CI debugging, thanks :)

jistr:
Trying 10->16 again

philroche (approved):

PR #452, opened 2021-07-28
Fix: in rhsm auto_attach once defaulted will always collide with acti…

ccamacho:
…vationkey

Once a variable is defaulted to an existing value it can not
be omited, auto_attach is defaulted to true so if activationkey
it can not be put back to omit.

omit is a special variable for omitting an individual argument
to a module with only that value, and once defined it can not
be omited again.

PR #453, opened 2021-07-29
Docs: Add password-based authentication variable docs.

ccamacho:
This commit adds the information for the ssh access using
password in the conversion hosts.

### DIFF
--- a/docs/src/user/variables-guide.rst
+++ b/docs/src/user/variables-guide.rst
@@ -196,6 +196,26 @@ Ansible module. In OS Migrate they are named as follows::
     os_migrate_conversion_rhsm_syspurpose
     os_migrate_conversion_rhsm_username
 
+Enabling password-based SSH access to the conversion hosts
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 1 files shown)


docs/src/user/variables-guide.rst
```diff
@@ -187,6 +196,26 @@ Ansible module. In OS Migrate they are named as follows::
     os_migrate_conversion_rhsm_syspurpose
     os_migrate_conversion_rhsm_username
 
+Enabling password-based SSH access to the conversion hosts
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
ccamacho, line 200: Ill rebase after #452 is merged, so the text in the prev lines is removed from this PR.

ccamacho: done

jistr:
Looks good, i'm just on the fence re the commit message whether it should be "Dev" or not, since we already do have a changelog entry for the password-based authentication under "New". If we leave the commit message as-is, it will appear in the "Other" section i think, and as the change affects user-facing docs, that's probably ok as this concerns user-facing docs after all. +1

jistr (approved):

PR #454, opened 2021-08-03
Fix: For resources that have name, reject migration with empty name

philroche:
For resources which have a name property validate that it is non empty.

This fixes an issue where resources with empty names are not included in the check
for uniqueness/idempotence. By checking for empty names we avoid this issue.

### DIFF
--- a/os_migrate/plugins/module_utils/resource.py
+++ b/os_migrate/plugins/module_utils/resource.py
@@ -238,6 +238,7 @@ class Resource():
         well-formed
         """
         errors = self._validation_id_errors()
+        errors.extend(self._validation_empty_name_errors())
         errors.extend(self._validation_migration_params_errors())
         errors.extend(self._validation_params_errors())
         return errors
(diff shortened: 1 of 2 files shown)


os_migrate/plugins/module_utils/resource.py
```diff
@@ -529,3 +530,20 @@ def _validation_params_errors(self):
...
+        Returns: list of strings (error messages)
+        """
+        errors = []
+        params = self.params()
+        name_param_key = 'name'
+        # check to see if this resource has a name
+        if name_param_key in self.params_from_sdk:
+            # Check to see if name is present and if so is it empty.
+            # _validation_params_errors check will catch if it is not present.
+            if name_param_key in params.keys() and not params.get(name_param_key, ''):
```
jistr, line 547: The condition could simply be `if params.get(name_param_key, ''):`. Not a blocker but may be worth amending?

jistr (approved):

os_migrate/plugins/module_utils/resource.py
jistr: This rather: `if not params.get(name_param_key):'

philroche: True but then "params.name is empty." will be appended to errors even when "name" is missing. I figured that this is not expected for this validation as "name" missing is handled by _validation_params_errors.

jistr: Actually now i realize why you did the `name_param_key in params.keys()` part -- to make sure we don't get duplicate validation errors for the same issue, right?  All good then, we can merge. Just will give some time for Carlos to have a chance to review as well.

philroche: @jistr exactly :)

jistr:
Some time has passed, merging.

PR #455, opened 2021-08-05
Dev: Reduce the container image size

ccamacho:
This commit helps to reduce the space burden
of the os-migrate container image.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 SHELL := /bin/bash
 ROOT_DIR := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
 OS_MIGRATE := $(HOME)/.ansible/collections/ansible_collections/os_migrate/os_migrate
-NO_VAGRANT := 0
+NO_VAGRANT ?= 0
 export OS_MIGRATE
 
 
--- a/docs/src/user/install-from-galaxy.rst
+++ b/docs/src/user/install-from-galaxy.rst
@@ -47,5 +47,5 @@ you can use virtualenv, e.g.::
 
    python3 -m venv $HOME/os_migrate_venv
    source $HOME/os_migrate_venv/bin/activate
-   pip install --upgrade 'openstacksdk>=0.36'
-   pip install --upgrade 'ansible>=2.9.1,<2.10'
+   python3 -m pip install --upgrade 'openstacksdk>=0.36'
+   python3 -m pip install --upgrade 'ansible>=2.9.1,<2.10'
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -2,7 +2,6 @@
 SHELL := /bin/bash
 ROOT_DIR := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
 OS_MIGRATE := $(HOME)/.ansible/collections/ansible_collections/os_migrate/os_migrate
-NO_VAGRANT := 0
```
jistr, line 5: Let's use `?=` instead of `:=` but keep defaulting the variable? Phil added this setting to get rid of a warning about an undefined variable.

docs/src/user/install-from-galaxy.rst
```diff
@@ -47,5 +47,5 @@ you can use virtualenv, e.g.::
 
    python3 -m venv $HOME/os_migrate_venv
    source $HOME/os_migrate_venv/bin/activate
-   pip install --upgrade 'openstacksdk>=0.36'
-   pip install --upgrade 'ansible>=2.9.1,<2.10'
+   python3 -m pip install --upgrade 'openstacksdk>=0.36'
+   python3 -m pip install --upgrade 'ansible>=2.9.1,<2.10'
```
jistr, line 51: Is there a reason behind this change? Is this something recommended or better in some way? (I don't mind the change if it still does the job, just would like to know the rationale.)

scripts/build.sh
```diff
@@ -18,3 +22,5 @@ sed -i -e "s/^OS_MIGRATE_VERSION = .*$/OS_MIGRATE_VERSION = '$VERSION'  # update
 ansible-galaxy collection build os_migrate -v --force --output-path releases/
 cd releases
 ln -sf os_migrate-os_migrate-$VERSION.tar.gz os_migrate-os_migrate-latest.tar.gz
+
+python3 -m pip cache purge
```
jistr, line 26: IMO this is not necessary since the above install only happens in a started container, not in the container image itself. But probably doesn't hurt either.

jistr (changes requested):

Makefile
philroche: my bad :(

jistr: No worries! :)

docs/src/user/install-from-galaxy.rst
ccamacho: yeah its something im starting to see more and more, when we use venvs, so when you use pip directly you dont know which interpreter you will be using in the otehr case you do. let me get you some ref I was reading before this change

ccamacho: i.e. this https://snarky.ca/why-you-should-use-python-m-pip/

jistr: :+1: Thanks :)

ccamacho (commented):
test

philroche (approved):

PR #456, opened 2021-08-05
New: Ability to specify boot volume params when source VM has no boot volume

### LINKED ISSUE #444: Need to set availability zone for destination volume
In my destination cloud there is no default cinder availability zone, and the source AZ doesn't exist in destination.

So we need a parameter to set destination AZ.

jistr:
plus small refactors

(Described in commit messages.)

Resolves #444

### DIFF
--- a/docs/src/user/migration-params-guide.rst
+++ b/docs/src/user/migration-params-guide.rst
@@ -31,6 +31,20 @@ Workload migration parameters
       already boot-from-volume in the source cloud, this is the
       default and the only possible path.)
 
+-  ``boot_volume_params`` controls new boot disk creation parameters in
+   the destination, in case the source VM has no boot disk, but
+   ``boot_disk_copy`` is ``true``.
+
+   There are a few pre-filled parameters, defaulted to ``None``.  Then
+   ``None`` value means those parameters will not be specified when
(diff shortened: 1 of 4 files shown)


philroche (commented):
Not exactly a query on this PR specifically but I can't find any docstrings/comments on what ser_server is. Is it an instance of an openstack server/vm?

jistr:
> Not exactly a query on this PR specifically but I can't find any docstrings/comments on what ser_server is. Is it an instance of an openstack server/vm?

`ser_` stands for serialized, so here it's a "serialized server". `ser_server` is an instance of the OS Migrate's Server class. The other idiom sometimes used in code would be `sdk_`, so `sdk_server` would be an instance of OpenStack SDK's Server class.

philroche:
> > Not exactly a query on this PR specifically but I can't find any docstrings/comments on what ser_server is. Is it an instance of an openstack server/vm?
> 
> `ser_` stands for serialized, so here it's a "serialized server". `ser_server` is an instance of the OS Migrate's Server class. The other idiom sometimes used in code would be `sdk_`, so `sdk_server` would be an instance of OpenStack SDK's Server class.

Thank you

philroche (approved):

PR #457, opened 2021-08-17
new: add a floating IP when creating the server instance

ccamacho:
This commit makes sure the conversion host have
assigned a floating IP after its creation.


ccamacho:
I tested 3 scenarios here:

scenario 1: No auto_ip and no floating_ip explicit allocation
    This fails as the conversion hosts are not reachable with each other, but it does not time out.
scenario 2: Yes auto_ip and floating_ip explicit allocation
    In the case, the instance creation succeeds, the floating_ip explicit allocation won't do anything as we have enabled the reuse option enabled.
scenario 3: No auto_ip and floating_ip explicit allocation
     In this case, the nova instance is created without a floating IP which is assigned in the next task.
     
We just benefit from the floating IP auto allocation when the external network defaults work which might not be feasible in almost any use case because by default we don't configure the external network where the floating IPs will be coming from. IIRC the pool where they come from default is called public, like the external network.

philroche (commented):
Can you add some context on why the change in timeout and also why this change is now required. Are floating IPs assigned by default in most environments?

ccamacho:
> Can you add some context on why the change in timeout and also why this change is now required. Are floating IPs assigned by default in most environments?

Yeah, when we deploy the conversion hosts the timeout was set to 10 minutes so we wait for the floating IP, in this case, we also wait the same amount of time when we allocate it, so we always wait 2 times os_migrate_conversion_host_deploy_timeout, and 600 seconds is just too much.

philroche (approved):

os_migrate/roles/conversion_host/tasks/main.yml
```diff
@@ -117,3 +117,20 @@
     ca_cert: "{{ os_migrate_conversion_ca_cert|default(omit) }}"
```
jistr, line 117: We need to set `auto_ip: no` above.

jistr: Also another thing that we wanted to support here was explicit assignment of pre-created FIPs (see the card description in Jira). So we could have something like `floating_ips: "{{ os_migrate_conversion_floating_ips|default(omit) }}"` here and on the higher level we'd have the `os_migrate_src_conversion_floating_ips` and `os_migrate_dst_conversion_floating_ips` split as we have with the other variables?

We could also make this a separate card and make a separate PR for this feature if we want? So this one is optional i think but the `auto_ip` change needs to be made at least.

os_migrate/roles/conversion_host/tasks/main.yml
```diff
@@ -117,3 +117,20 @@
...
+    wait: yes
+    timeout: "{{ os_migrate_conversion_host_deploy_timeout }}"
+    reuse: yes
+    auth: "{{ os_migrate_conversion_auth }}"
+    auth_type: "{{ os_migrate_conversion_auth_type|default(omit) }}"
+    region_name: "{{ os_migrate_conversion_region_name|default(omit) }}"
+    validate_certs: "{{ os_migrate_conversion_validate_certs|default(omit) }}"
+    ca_cert: "{{ os_migrate_conversion_ca_cert|default(omit) }}"
+    client_cert: "{{ os_migrate_conversion_client_cert|default(omit) }}"
+    client_key: "{{ os_migrate_conversion_client_key|default(omit) }}"
```
jistr, line 136: If we go for the manual floating IPs, this task would then probably have `when: os_migrate_conversion_floating_ips is defined and os_migrate_conversion_floating_ips not is string`.

(I think defaulting to `omit` on the `src/dst`-specific variables generates a value which is some specially formed string, so we need to check that the value is not a string. I wanted to check whether the value is list, there are `is iterable` and `is sequence` tests in Jinja, however they both match both a string and a list. There does not seem to be a Jinja test to specifically check for list only, based on my googling. :shrug:

jistr (changes requested):

os_migrate/roles/conversion_host/tasks/main.yml
jistr: Reference: https://jon.sprig.gs/blog/post/1801

jistr:
> I tested 3 scenarios here:
> 
> scenario 1: No auto_ip and no floating_ip explicit allocation
> This fails as the conversion hosts are not reachable with each other, but it does not time out.
> scenario 2: Yes auto_ip and floating_ip explicit allocation
> In the case, the instance creation succeeds, the floating_ip explicit allocation won't do anything as we have enabled the reuse option enabled.
> scenario 3: No auto_ip and floating_ip explicit allocation
> In this case, the nova instance is created without a floating IP which is assigned in the next task.
> 
> We just benefit from the floating IP auto allocation when the external network defaults work which might not be feasible in almost any use case because by default we don't configure the external network where the floating IPs will be coming from. IIRC the pool where they come from default is called public, like the external network.

But you went into code to edit the auto_ip for attempts, right?

The current state of the pull request doesn't fix the problem, because one of these will happen IIUC:

1. `auto_ip: yes` and it succeeds, so `floating_ip` won't do anything.
2. `auto_ip: yes` and it fails, so the `server` module fails with it, and the `floating_ip` is never executed.

So to fix i think we need `auto_ip: no` so that we avoid the auto_ip bug, and `floating_ip` takes care of the FIPs. And ideally also support the explicit providing of floating_ips as the card suggested, but we can split this out into a separate card+PR if desired.

os_migrate/roles/conversion_host/tasks/main.yml
ccamacho: both are done, I didn't add this change at first because I didn't (don't) know how to test it in our CI currently.

os_migrate/roles/conversion_host/tasks/main.yml
ccamacho: Yeah the problem is that  floating_ip_address is a string so we should default it to something not a string, like false

os_migrate/roles/conversion_host/tasks/main.yml
ccamacho: another thing I didnt like is to duplicate the task

os_migrate/roles/conversion_host/tasks/main.yml
ccamacho: > If we go for the manual floating IPs, this task would then probably have `when: os_migrate_conversion_floating_ips is defined and os_migrate_conversion_floating_ips not is string`.
> 
> (I think defaulting to `omit` on the `src/dst`-specific variables generates a value which is some specially formed string, so we need to check that the value is not a string. I wanted to check whether the value is list, there are `is iterable` and `is sequence` tests in Jinja, however they both match both a string and a list. There does not seem to be a Jinja test to specifically check for list only, based on my googling. 🤷 )

I'm afraid that `is defined` will always pass and the parameter must be a string, in this case if we default it to false, we can check whether if the variable is a string or not as it should be a string (https://docs.ansible.com/ansible/latest/collections/openstack/cloud/floating_ip_module.html#parameter-floating_ip_address)

docs/src/user/variables-guide.rst
```diff
@@ -116,6 +116,17 @@ The variables to be configured are::
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Each conversion host needs to have a floating IP,
+this floating IP can be assigned automatically or
+defined by the operator with the usage of the
+following variables::
+
+    os_migrate_src_conversion_fixed_floating_ip_address    
+    os_migrate_dst_conversion_fixed_floating_ip_address
+
```
jistr, line 129: I think the naming pattern should be just `os_migrate_src_conversion_floating_ip_address` (and dst respectively). The term "fixed IP address" is a differently defined term in OpenStack, so we should avoid confusing users here.

os_migrate/roles/conversion_host/tasks/main.yml
```diff
@@ -117,3 +117,40 @@
     ca_cert: "{{ os_migrate_conversion_ca_cert|default(omit) }}"
     client_cert: "{{ os_migrate_conversion_client_cert|default(omit) }}"
     client_key: "{{ os_migrate_conversion_client_key|default(omit) }}"
+
+- name: create the conversion host floating IP with a fixed IP
```
jistr, line 121: s/fixed/predefined/ ?

os_migrate/roles/conversion_host/tasks/main.yml
```diff
@@ -117,3 +117,40 @@
...
+  openstack.cloud.floating_ip:
+    server: "{{ os_migrate_conversion_host_name }}"
+    network: "{{ os_migrate_conversion_external_network_name }}"
```
jistr, line 124: The network parameter should be removed from the case with floating_ip_address, they are mutually exclusive. I looked up the code and if `floating_ip_address` is specified, the module itself will drop `network`, but i think if we do another edit here, it would be nicer to drop the parameter here and not rely on the module doing it. (It could just as well behave the other way - ignore floating_ip_address if network is provided, we are just lucky here.)

https://github.com/openstack/ansible-collections-openstack/blob/ae4e7f3c0607ccf2725b25a6d7d321eb3c32e6be/plugins/modules/floating_ip.py#L269

jistr (changes requested):

os_migrate/playbooks/deploy_conversion_hosts.yml
```diff
@@ -20,6 +20,7 @@
         os_migrate_conversion_net_mtu: "{{ os_migrate_src_conversion_net_mtu|default(omit) }}"
         os_migrate_conversion_subnet_dns_nameservers:
           "{{ os_migrate_src_conversion_subnet_dns_nameservers|default(omit) }}"
+        os_migrate_conversion_fixed_floating_ip_address: "{{ os_migrate_src_conversion_floating_ip_address|default(false) }}"
```
jistr, line 23: Should we fix the inner name too, to keep consistency in the variable naming?

docs/src/user/variables-guide.rst
ccamacho: done

os_migrate/roles/conversion_host/tasks/main.yml
ccamacho: done

os_migrate/roles/conversion_host/tasks/main.yml
ccamacho: done

os_migrate/playbooks/deploy_conversion_hosts.yml
ccamacho: yeah lets have it all the same, done

jistr (approved):

os_migrate/playbooks/deploy_conversion_hosts.yml
jistr: Thanks :)

jistr:
Needs a rebase :construction:

jistr:
I wonder if the CI failures are caused by the networking resources no longer being properly cleaned up as we use the `floating_ip` module explicitly (floating IPs not autodeleted?). Maybe we need to make sure we explicitly delete the conversion hosts' floating IPs?

ccamacho:
no, @jistr this means that when the OSP10->OSP13 migration fails and the resources are not cleaned up correctly, when we deploy with a different name (10->16) the automation wouldn't know which resources are failing.

PR #458, opened 2021-08-18
new: allow configuring the conversion hosts names

ccamacho:
This commit allows to configure the conversion hosts
names.


ccamacho:
We are testing this by setting custom names in the conversion hosts for the 10 -> 16 job and leaving the default values for the 13 -> 16 job so we cover both scenarios, operators configuring them or not.

ccamacho:
In general, I think this should be a regression as I implemented this change in the past but it got removed from the collection.

PR #459, opened 2021-08-18
new: allow running scripts in the conversion hosts

ccamacho:
This commit allows to run custom bash scripts in
the conversion hosts before and after the content
playbook.

### DIFF
--- a/os_migrate/playbooks/deploy_conversion_hosts.yml
+++ b/os_migrate/playbooks/deploy_conversion_hosts.yml
@@ -79,7 +79,7 @@
     ansible_become_user: root
     ansible_become: true
 
-- hosts: conversion_hosts
+- hosts: os_migrate_conv_src
   any_errors_fatal: true
   tasks:
     - name: install the conversion host content
@@ -89,3 +89,19 @@
   vars:
     ansible_become_user: root
     ansible_become: true
+    os_migrate_conversion_host_pre_content_hook: "{{ os_migrate_src_conversion_host_pre_content_hook|default(false) }}"
+    os_migrate_conversion_host_post_content_hook: "{{ os_migrate_src_conversion_host_post_content_hook|default(false) }}"
+
+
+- hosts: os_migrate_conv_dst
+  any_errors_fatal: true
+  tasks:
+    - name: install the conversion host content
+      ansible.builtin.include_role:
+        name: os_migrate.os_migrate.conversion_host_content
+      when: os_migrate_conversion_host_content_install|default(true)|bool
+  vars:
+    ansible_become_user: root
+    ansible_become: true
+    os_migrate_conversion_host_pre_content_hook: "{{ os_migrate_dst_conversion_host_pre_content_hook|default(false) }}"
+    os_migrate_conversion_host_post_content_hook: "{{ os_migrate_dst_conversion_host_post_content_hook|default(false) }}"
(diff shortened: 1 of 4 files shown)


os_migrate/roles/conversion_host_content/tasks/main.yml
```diff
@@ -6,10 +6,16 @@
           Conversion Host OS is
           {{ ansible_distribution }} {{ ansible_distribution_version }}
 
+    - name: running pre content hook
+      ansible.builtin.shell: "{{ os_migrate_conversion_host_pre_content_hook | default(omit) }}"
```
ccamacho, line 10: im thinking about if this can be a blockinfile and then we render and execute the content

os_migrate/roles/conversion_host_content/tasks/main.yml
```diff
@@ -6,10 +6,18 @@
           Conversion Host OS is
           {{ ansible_distribution }} {{ ansible_distribution_version }}
 
+    - name: running pre content hook
+      ansible.builtin.shell: "{{ os_migrate_conversion_host_pre_content_hook }}"
```
ccamacho, line 10: im thinking about if this can be a blockinfile and then we render and execute the content

ccamacho:
Tested with these variables, so we test single-line and multi-line commands and we also leave undefined some.

```
if [[ "$SRC_CLOUD" == "src10" && "$DST_CLOUD" == "dst16" ]]; then
cat << EOF > tests/e2e/tenant/scenario_variables.yml
os_migrate_dst_conversion_host_post_content_hook: "echo 'single command'"
os_migrate_src_conversion_host_pre_content_hook: |
    ls -ltah
    echo "hello world"
    df -h
EOF
fi

if [[ "$SRC_CLOUD" == "src13" && "$DST_CLOUD" == "dst16" ]]; then
cat << EOF > tests/e2e/tenant/scenario_variables.yml
os_migrate_src_conversion_host_post_content_hook: "echo 'single command'"
os_migrate_dst_conversion_host_pre_content_hook: |
    ls -ltah
    echo "hello world"
    df -h
EOF
fi
```

philroche (approved):

os_migrate/playbooks/deploy_conversion_hosts.yml
```diff
@@ -79,7 +79,21 @@
...
+        name: os_migrate.os_migrate.conversion_host_content
+      when: os_migrate_conversion_host_content_install|default(true)|bool
+  vars:
+    ansible_become_user: root
+    ansible_become: true
+    os_migrate_conversion_host_pre_content_hook: "{{ os_migrate_src_conversion_host_pre_content_hook|default(false) }}"
+    os_migrate_conversion_host_post_content_hook: "{{ os_migrate_src_conversion_host_post_content_hook|default(false) }}"
+
+
+- hosts: os_migrate_conv_dst
```
jistr, line 96: This is going to slow down the conversion host deployment, because until now, the installation of content went on in parallel, but this is going to make it go in series. I don't know of a better solution though, given that we have the variable name split. And the variable name split makes sense here. So +1

os_migrate/roles/conversion_host_content/tasks/main.yml
jistr: Perhaps, but i think what we have here is good enough. Typically folks run Ansible with `-v` so some detail would be shown this way too.

jistr (approved):

PR #461, opened 2021-08-18
new: allow disabling the RHSM tasks

ccamacho:
This commit allows to disable the execution
of the RHSM tasks.


philroche (commented):
LGTM but can we add some detail as to why this is required?

ccamacho:
> LGTM but can we add some detail as to why this is required?

Yeah, for example, if the operator has a custom rhsm script, ideally they should be able to run it with the bash hooks and then disabling this block so there is a mutual exclusive mechanism to choose where to run rhsm from.

philroche (approved):
Thanks

ccamacho:
I added the `os_migrate_conversion_rhsm_manage` variable to false and the deployment failed as the rhsm tasks were not executed.

philroche (approved):

jistr (changes requested):
We should also use the `_manage` variable at https://github.com/os-migrate/os-migrate/blob/cb3aa0f450dac3dc7bbd09b05ab39fa2d57419a6/os_migrate/roles/conversion_host/tasks/unregister.yml#L8

Otherwise LGTM, thanks!

ccamacho:
> We should also use the `_manage` variable at
> 
> https://github.com/os-migrate/os-migrate/blob/cb3aa0f450dac3dc7bbd09b05ab39fa2d57419a6/os_migrate/roles/conversion_host/tasks/unregister.yml#L8
> 
> Otherwise LGTM, thanks!

Done!

ccamacho:
Blocked by: https://github.com/os-migrate/os-migrate/pull/463

os_migrate/roles/conversion_host/tasks/unregister.yml
```diff
@@ -6,5 +6,6 @@
     ansible_become: true
     ansible_become_user: root
   when:
+    - os_migrate_conversion_rhsm_manage and
```
jistr, line 9: I think the `and` at the end will break it?

jistr: These are separate list items with an implicit `and` operation between them.

ccamacho: right, sorry I copied that from the other condition and didn't remove the and this will fails for sure

os_migrate/playbooks/delete_conversion_hosts.yml
```diff
@@ -15,8 +15,8 @@
   tasks:
     - name: unregister the conversion hosts
       ansible.builtin.include_role:
-        name: os_migrate.os_migrate.conversion_host
-        tasks_from: unregister.yml
+        name: os_migrate.os_migrate.conversion_host_content
```
ccamacho, line 18: We should have these tasks as part of the same role, I believed we had it that way

jistr: +1 good catch

jistr:
This should be merged before #457 is being rebased to reduce the chance of further conflicts. (There is already a conflict on #457 with a different PR.)

PR #464, opened 2021-08-23
New: Conditionally stop VMs before migration

philroche:
Set os_migrate_workload_stop_before_migration to true to stop VMs before migration occurs.


ccamacho:
Retrying 10->16: Unable to reach the server at subscription.rhn.stage.redhat.com:443/subscription

ccamacho:
Testing with: 
```
if [[ "$SRC_CLOUD" == "src10" && "$DST_CLOUD" == "dst16" ]]; then
cat << EOF > tests/e2e/tenant/scenario_variables.yml
os_migrate_workload_stop_before_migration: true
EOF
fi
```

os_migrate/roles/import_workloads/defaults/main.yml
```diff
@@ -4,5 +4,6 @@ os_migrate_conversion_keypair_private_path: "{{ os_migrate_data_dir }}/conversio
 os_migrate_timeout: 1800
 os_migrate_workload_cleanup_on_failure: true
 os_migrate_workload_boot_volume_prefix: os-migrate-
+os_migrate_workload_stop_before_migration: false
```
ccamacho, line 7: This should be documented

os_migrate/roles/import_workloads/tasks/workload.yml
```diff
@@ -34,6 +34,51 @@
...
+      region_name: "{{ os_migrate_src_region_name|default(omit) }}"
+      validate_certs: "{{ os_migrate_src_validate_certs|default(omit) }}"
+      ca_cert: "{{ os_migrate_src_ca_cert|default(omit) }}"
+      client_cert: "{{ os_migrate_src_client_cert|default(omit) }}"
+      client_key: "{{ os_migrate_src_client_key|default(omit) }}"
+      timeout: "{{ os_migrate_timeout }}"
+      wait: yes
+      filters:
+        vm_state: STOPPED
+    register: _workload_info
```
ccamacho, line 70: I like this pattern to register vars with the _ prefix

os_migrate/roles/import_workloads/defaults/main.yml
ccamacho: About `os_migrate_workload_stop_before_migration` do we want to only allow stopping VMs before the migration? if we do this we will be changing the operator "regular behavior", do we want to allow them to start those stopped VMs after?

When stopping the VMs is useful? When the VMs have state, i.e. a DB server in that case is useful to stop/migrate but then the apps will fails as the DB server is stopped.

Something like `os_migrate_workload_start_after_migration` I think will be a nice addition, we could have only the "stop-only" and a follow up PR to handle this logic.

ccamacho:
Nice, it looks now that the CI is testing this PR correctly, everytime you add the 10->16 label this feature will be tested. So feel free to force-push the branch and add the label to test it.

os_migrate/roles/import_workloads/tasks/workload.yml
```diff
@@ -34,6 +34,51 @@
...
+      validate_certs: "{{ os_migrate_src_validate_certs|default(omit) }}"
+      ca_cert: "{{ os_migrate_src_ca_cert|default(omit) }}"
+      client_cert: "{{ os_migrate_src_client_cert|default(omit) }}"
+      client_key: "{{ os_migrate_src_client_key|default(omit) }}"
+    when:
+      - prelim.changed
+      - os_migrate_workload_stop_before_migration
+
+  # Confirm that VM is now is expected state
+  - name: gather info on stopped workload
```
jistr, line 56: Whether the VM is indeed stopped will be checked by the `import_workload_src_check` module, so we don't need to double that i think.

os_migrate/roles/import_workloads/tasks/workload.yml
```diff
@@ -34,6 +34,51 @@
       dst_filters: "{{ os_migrate_dst_filters }}"
     when: prelim.changed
 
+  # Stop VM if not stopped and if config set to true
+  - name: perform workload stop if os_migrate_workload_stop_before_migration is true
+    openstack.cloud.server_action:
+      auth: "{{ os_migrate_src_auth }}"
+      auth_type: "{{ os_migrate_src_auth_type|default(omit) }}"
+      region_name: "{{ os_migrate_src_region_name|default(omit) }}"
+      server: "{{ item }}"
```
jistr, line 43: `item` is the serialized data structure (complex dict) of the server. We should use the ID here, so `server: "{{ item['_info']['id'] }}"`

os_migrate/roles/import_workloads/defaults/main.yml
jistr, line 7: +1 for defaulting this to `false` in production. However, we should default to `true` in tests, here:

https://github.com/os-migrate/os-migrate/blob/92c359e4e726ec53f2c0a519c781917bbe60cb68/tests/e2e/tenant/scenario_variables.yml

And we should remove this:

https://github.com/os-migrate/os-migrate/blob/92c359e4e726ec53f2c0a519c781917bbe60cb68/tests/e2e/tenant/seed_workload.yml#L33-L43

That way your new code here will be tested by the e2e tests/CI.

jistr (changes requested):

os_migrate/roles/import_workloads/defaults/main.yml
jistr: +1 for docs, but for now i'd be -1 on `start_after_migration`. I think that once the workload is migrated, we want it running in the destination cloud, rather than running it duplicately? There might be some special cases where duplicating is the desired case but IMO let's discuss that separately + create a separate task for that if we indeed want it.

philroche: Docs added in https://github.com/os-migrate/os-migrate/pull/464/commits/2a310a2820827fe8651e9867d6e2eb2c7d5c8859

philroche: Thank you. Changes made in https://github.com/os-migrate/os-migrate/pull/464/commits/96bed12fd77332717c5c0e5c4c35f13e5f1a900f and https://github.com/os-migrate/os-migrate/pull/464/commits/8434ba1fd306dafbbcd9e912762afee7cc4f143c

os_migrate/roles/import_workloads/tasks/workload.yml
philroche: Thanks. Changes made in https://github.com/os-migrate/os-migrate/pull/464/commits/f551ffb9d0f5972b88eea01ddabeebc97421cb77

os_migrate/roles/import_workloads/tasks/workload.yml
philroche: Thanks - I had seen that in my testing too. Changes made in https://github.com/os-migrate/os-migrate/pull/464/commits/f57d4d2b7c7275ffc061610a38aaef13bdf0c4d6

ccamacho:
@philroche can you squash all these changes together?

philroche:
I will yes but I wanted to have them separate now so you could see the requested changes.

PR #472, opened 2021-08-31
Fix: Validate keypairs exist in destination cloud before migrating workload

philroche:
dst_prerequisites_errorsclass method is used to validate prerequisites
before migrating workloads but the parent class dst_prerequisites_errors
does not validate keypairs exist in destination cloud. To do this we do
a lookup to see if keypair name exists and append to errors list if not

### DIFF
--- a/os_migrate/plugins/module_utils/server.py
+++ b/os_migrate/plugins/module_utils/server.py
@@ -248,3 +248,22 @@ class Server(resource.Resource):
         refs['volumes'] = params['volumes']
 
         return refs
+
+    def dst_prerequisites_errors(self, conn, filters=None):
+        """Get messages for unmet destination cloud prerequisites including keypairs.
+
+        Returns: list of strings (error messages), empty when prerequisites
+        are met
+        """
+        errors = super().dst_prerequisites_errors(conn, filters)
+
+        # Parent class dst_prerequisites_errors does not validate keypairs exist in destination cloud.
+        # To do this we do a lookup to see if keypair name exists and append to errors list if not
+        params = self.params()
+        try:
+            conn.compute.find_keypair(params['key_name'], ignore_missing=False)
+        except (openstack.exceptions.ResourceFailure,
+                openstack.exceptions.ResourceNotFound,
+                openstack.exceptions.DuplicateResource) as e:
+            errors.append("Destination keypair prerequisites not met: {0}".format(e))
+        return errors
--- a/os_migrate/tests/unit/test_server.py
+++ b/os_migrate/tests/unit/test_server.py
@@ -6,6 +6,7 @@ import unittest
 
 from ansible_collections.os_migrate.os_migrate.plugins.module_utils \
     import server
+from ansible_collections.os_migrate.os_migrate.tests.unit.test_keypair import sdk_keypair
 
 
 def sdk_server():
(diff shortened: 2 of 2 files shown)


os_migrate/plugins/module_utils/server.py
```diff
@@ -248,3 +248,22 @@ def _refs_from_ser(self, conn, filters=None):
...
+        Returns: list of strings (error messages), empty when prerequisites
+        are met
+        """
+        errors = super().dst_prerequisites_errors(conn, filters)
+
+        # Parent class dst_prerequisites_errors does not validate keypairs exist in destination cloud.
+        # To do this we do a lookup to see if keypair name exists and append to errors list if not
+        params = self.params()
+        try:
+            conn.compute.find_keypair(params['key_name'], **filters)
```
jistr, line 264: Actually we need to pass `ignore_missing=False`, which the reference lookup methods in the parent class do. Otherwise the method returns None rather than raising an exception.

jistr (changes requested):

os_migrate/plugins/module_utils/server.py
jistr: https://docs.openstack.org/openstacksdk/latest/user/proxies/compute.html#keypair-operations

jistr: https://github.com/os-migrate/os-migrate/blob/d86c0a27fcff1cc208ebe2b244993bfba5cabd22/os_migrate/plugins/module_utils/reference.py#L374-L378

philroche: Thanks.

os_migrate/plugins/module_utils/server.py
```diff
@@ -248,3 +248,24 @@ def _refs_from_ser(self, conn, filters=None):
...
+        """
+        errors = super().dst_prerequisites_errors(conn, filters)
+
+        # Parent class dst_prerequisites_errors does not validate keypairs exist in destination cloud.
+        # To do this we do a lookup to see if keypair name exists and append to errors list if not
+        params = self.params()
+        try:
+            # find_keypair only accepts user_id as a filter. Check if user_id in filters dict else use None
+            user_id = filters.get('user_id', None)
+            conn.compute.find_keypair(params['key_name'], ignore_missing=False, user_id=user_id)
```
jistr, line 266: Not sure if providing `user_id=None` is the same as not providing it at all, but we'll see in the e2e tests i think.

More generally, unless i'm mistaken, we actually stopped using the filters for destination-side lookups as the references carry their own project/domain info nowadays. We just pass around the destination filters because they were used previously. Later we may want to check whether it's indeed the case, and remove the filters from destination-side lookups everywhere.

This is just for context though. If the CI passes, i think we can merge like this.

jistr: 13->16 is green

jistr (approved):

os_migrate/plugins/module_utils/server.py
philroche: Ack.

The default for user_id kwarg for find_keypair is None and it is checked

`attrs = {'user_id': user_id} if user_id else {}`

philroche: it appears that there are versions of openstacksdk which do not accept user_id 

`find_keypair() got an unexpected keyword argument 'user_id'` from https://gitlab.cee.redhat.com/os-migrate/bot/-/jobs/4416270

Removing user_id as you suggested.

jistr: Looks good.

jistr (approved):
Looks good, thanks!

PR #477, opened 2021-09-03
new: allow deleting or detaching the conversion hosts FIPs

ccamacho:
This commit enables users to detach or demove floating IPs
when deleting the conversion hosts.


jistr (approved):

ccamacho:
10 -> 16 unrelated error:   msg: Unable to reach the server at subscription.rhn.stage.redhat.com:443/subscription

PR #484, opened 2021-09-10
Dev: Prepare os-migrate for ansible version bump in toolbox build to latest known working version 4.5.0

philroche:
ansible 4.5.0 is known to work for all e2e and functional tests so we can now
bump the ansible version installed in the toolbox virtualenv to 4.5.0


jistr:
All CI green :+1: . I'll try to review the whole and then we can think how to squash (one commit, multiple commits, multiple PRs).

@ccamacho +1 on the reviewing not being easy here, but should the splitting give @philroche additional headache, personally i'd be ok with keeping a single PR, and maybe splitting into 2 commits? One for the bump of Ansible version and the uninstall mechanism for requirement overrides, and another commit for all linter fixes? The linter commit would probably be 'Dev' and the Ansible commit is relevant to changelog (users might be interested we started testing with newer Ansible), so it could be New/Chg/Oth?

WDYT?

ccamacho:
+1 Next time we should try, having different PRs from the beginning. We don't have any other PR blocked so let's say this PR make the changes and fixes the CI itself, I think at least 2 commits should be here, the CI fixes and the ansible version change. @philroche thanks for these improvements!!!! @jistr about the ansible version we ship? Do you think we should always deploy latest Ansible? What if this is deployed outside the toolbox container? If using rhel then the version will be 2.9.x which might break. WDYT?

philroche:
@ccamacho I have a follow on task to add a matrix to our CI/github actions to test the three major supported ansible versions - 2.9.X, 3.4.0 & 4.5.0. Would that being in place ease your concerns about the shipped version int he toolbox?

ccamacho:
@philroche amazing!!!!! Thanks 🙏🙏🙏🙏 then +1 from my side

jistr:
Ah yes i forgot to comment on that. +1 on the problem raised by Carlos and +1 on Phil's suggestion there. We do need to keep testing 2.9 as we need to keep it working for the reasons @ccamacho outlined, but i think we should have another set of jobs which aims to move forward with upstream Ansible.

We should have done it in reverse order, first expand the matrix to multiple jobs (running the same thing), then changing the Ansible version in one of them.

What about merging all the changes *except* the Ansible bump in requirements.txt, as the refactors should keep working with 2.9 still. And then when we have the matrix, just edit requirements.txt in one job to start testing with newer Ansible, and hopefully all the other necessary changes would already be in place, so it would be a very small commit?

ccamacho:
@jistr agreed :)

os_migrate/plugins/modules/import_workload_dst_failure_cleanup.py
```diff
@@ -231,9 +237,10 @@ def _delete_volumes(self):
 
 def run_module():
     argument_spec = openstack_full_argument_spec(
+        auth=dict(type='dict', no_log=True, required=True),
         data=dict(type='dict', required=True),
         conversion_host=dict(type='dict', required=True),
-        ssh_key_path=dict(type='str', required=True),
+        ssh_key_path=dict(type='str', required=True, no_log=True),
```
jistr, line 243: It's just a path, not the actual content of the key. I don't think this should be no_log, we may want this as debugging info, but it should not cause any security concerns even when logged IMO.

os_migrate/plugins/modules/import_workload_export_volumes.py
```diff
@@ -517,10 +522,11 @@ def _export_volumes_from_converter(self):
 
 def run_module():
     argument_spec = openstack_full_argument_spec(
+        auth=dict(type='dict', no_log=True, required=True),
         data=dict(type='dict', required=True),
         boot_volume_prefix=dict(type='str', default=None),
         conversion_host=dict(type='dict', required=True),
-        ssh_key_path=dict(type='str', required=True),
+        ssh_key_path=dict(type='str', required=True, no_log=True),
```
jistr, line 529: Ditto

os_migrate/plugins/modules/import_workload_src_cleanup.py
```diff
@@ -405,9 +411,10 @@ def _attach_data_volumes_to_source(self):
 
 def run_module():
     argument_spec = openstack_full_argument_spec(
+        auth=dict(type='dict', no_log=True, required=True),
         data=dict(type='dict', required=True),
         conversion_host=dict(type='dict', required=True),
-        ssh_key_path=dict(type='str', required=True),
+        ssh_key_path=dict(type='str', required=True, no_log=True),
```
jistr, line 417: Ditto

os_migrate/plugins/modules/import_workload_transfer_volumes.py
```diff
@@ -557,9 +563,10 @@ def _detach_destination_volumes(self):
 
 def run_module():
     argument_spec = openstack_full_argument_spec(
+        auth=dict(type='dict', no_log=True, required=True),
         data=dict(type='dict', required=True),
         conversion_host=dict(type='dict', required=True),
-        ssh_key_path=dict(type='str', required=True),
+        ssh_key_path=dict(type='str', required=True, no_log=True),
```
jistr, line 569: Ditto

toolbox/build/build.sh
```diff
@@ -10,7 +10,6 @@ ANSIBLE_PYTHON=python3.8
 
 dnf -y update
 dnf -y install \
-    ansible \
```
jistr, line 13: This won't break CI, but it will break the Vagrant provisioning, won't it?

jistr (changes requested):
Some comments inline, other than that it looks good, and the more i think about it, the more i think we should indeed do the final Ansible switch only after we have the matrix. But the bulk of the PR can still be merged now.

toolbox/build/build.sh
philroche: it is still installed later in the 

```
if [ "${NO_VAGRANT:-0}" != "1" ]; then
```

clause

os_migrate/plugins/modules/import_workload_transfer_volumes.py
philroche: Now reverted in https://github.com/os-migrate/os-migrate/pull/484/commits/dcb6ba052d85e887f34570d66b489a6b7a0a09da

os_migrate/plugins/modules/import_workload_src_cleanup.py
philroche: Now reverted in https://github.com/os-migrate/os-migrate/pull/484/commits/dcb6ba052d85e887f34570d66b489a6b7a0a09da

os_migrate/plugins/modules/import_workload_export_volumes.py
philroche: Now reverted in https://github.com/os-migrate/os-migrate/pull/484/commits/dcb6ba052d85e887f34570d66b489a6b7a0a09da

os_migrate/plugins/modules/import_workload_dst_failure_cleanup.py
philroche: Now reverted in https://github.com/os-migrate/os-migrate/pull/484/commits/dcb6ba052d85e887f34570d66b489a6b7a0a09da

toolbox/build/build.sh
jistr: :+1: thanks

Makefile
```diff
@@ -136,7 +136,14 @@ test-sanity: reinstall
 		source /root/venv/bin/activate; \
 	fi; \
 	cd /root/.ansible/collections/ansible_collections/os_migrate/os_migrate; \
-	ansible-test sanity --skip-test import
+    # Only target the version of python3 in the activate virtual environment. Do not target; \
+    # any other discovered python runtimes.; \
+    TARGET_PYTHON_RUNTIME=$$(which python3); \
+    TARGET_PYTHON_VERSION=$$(python3 -c 'import sys; print(f"{sys.version_info.major}.{sys.version_info.minor}")'); \
+	echo "Running Ansible sanity tests with $${TARGET_PYTHON_RUNTIME} python interpreter \
```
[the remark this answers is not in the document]
philroche: I found it useful to see during debugging but I can remove

philroche:
How squashed to two commits - I will full retest before adding one final commit reverting the ansible version bump

philroche:
> Now squashed to two commits - I will full retest before adding one final commit reverting the ansible version bump

All tests passed with squashed commits. I will now revert the ansible version bump and test again

Makefile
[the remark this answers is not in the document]
jistr: ack :+1:

jistr (approved):

philroche:
> > Now squashed to two commits - I will full retest before adding one final commit reverting the ansible version bump
> 
> All tests passed with squashed commits. I will now revert the ansible version bump and test again

All tests have now passed with the ansible version bump reversed

toolbox/build/venv-requirements.txt
```diff
@@ -1,5 +1,5 @@
 # runtime
-ansible>=4.5.0,<4.6
+ansible>=2.9.1,<2.10
```
ccamacho: This should not be a revert, this should be a change in the original commit, is a revert once the code is merged, and this is not merged yet. Otherwise LGTM.

ccamacho (changes requested):

toolbox/build/venv-requirements.txt
ccamacho: We should have only 2 commits, not 3.

philroche: Done and PR title changed too

ccamacho (approved):

PR #490, opened 2021-09-24
Dev: Add ansible versions to github actions matrix config

philroche:
Dev: Add ansible versions to github actions matrix config

We can then test multiple ansible versions per PR submission


philroche:
Functional tests currently blocked on devstack install issues which @ccamacho is addressing in https://github.com/os-migrate/os-migrate/pull/492

philroche:
There are branch protection rules that require "build (3.7) " and "functional (3.7) " pass before PR can be merged. In this PR we are however changing the names of the tests as the matrix items are included in the name. 

If we are happy with this change then those branch protection rules need to be changed to include all three build tests and all three functional tests.

ccamacho:
++ on the change in the protection rules, we could after this PR runs, update the protection rules to use the new names, on every run, the protection rule names gets updated, so we could update/merge once it runs.

This change is awesome!

ccamacho:
Just squash this to a single commit and I think is good to go.

toolbox/build/venv-requirements-ansible-2.9.txt
```diff
@@ -0,0 +1 @@
+ansible>=2.9.1,<2.10
```
jistr, line 1: All looks good, i'm just wondering how to test the lower-bound supported OpenStack SDK as well. We could add the `openstacksdk` condition here. Or create another job for it in a separate PR?

Overall +1 that we can merge after squashing.

ccamacho:
I suggest to change the commit message like:

Dev: Add ansible versions to github actions matrix config

This commit includes:

- Allowing to test multiple versions of ansible in the
github actions matrix.

- Install required venv packages.

- Updating ansible-lint checks depending on the Ansible version.

philroche:
> I suggest to change the commit message like:
> 
> Dev: Add ansible versions to github actions matrix config
> 
> This commit includes:
> 
> * Allowing to test multiple versions of ansible in the
>   github actions matrix.
> * Install required venv packages.
> * Updating ansible-lint checks depending on the Ansible version.

Thanks. Now squashed

toolbox/build/venv-requirements-ansible-2.9.txt
philroche: I can add a new PR for that

jistr (approved):

PR #491, opened 2021-09-27
dev: make ansible-lint changes for latest

ccamacho:
This commit starts to include the latest changes to
support newer versions of ansible lint.


ccamacho:
Some context for this change: https://github.com/ansible-community/ansible-lint/issues/1398

os_migrate/localhost_inventory.yml
```diff
@@ -1,5 +0,0 @@
-migrator:
-  hosts:
-    localhost:
-      ansible_connection: local
-      ansible_python_interpreter: "{{ ansible_playbook_python }}"
```
ccamacho, line 5: My only concern https://github.com/os-migrate/os-migrate/commit/c6070287f57ca4383a945a54eb03a4b8acac37ec

How do we have variables in the inventory files? How does this gets populated?

PR #494, opened 2021-09-30
New: Allow using pre-existing floating IPs for migrated servers

jistr:
Two new floating_ip_mode migration parameter values are introduced:
'new' and 'existing'. The modes now behave this way:

* 'skip' - Do not create any floating IPs.

* 'new' - Create a new floating IP (auto-assigned address).

* 'existing' - Assume the floating IP address specified in the
  workload serialization is already assigned to the destination
  project, but not attached. Attach this floating IP. If this is not
  possible for some reason, fail.

* 'auto' - Attempt the 'existing' method of floating IP assignment,
  but instead of failing, fall back to the 'new' method.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ install: os_migrate-os_migrate-latest.tar.gz
 		source /root/venv/bin/activate; \
 	fi; \
 	cd releases; \
-	ansible-galaxy collection install --force os_migrate-os_migrate-latest.tar.gz
+	ansible-galaxy collection install $(OS_MIGRATE_INSTALL_ARGS) --force os_migrate-os_migrate-latest.tar.gz
 
 clean:
 	ls releases/os_migrate-os_migrate*.tar.gz | xargs rm
--- a/os_migrate/plugins/module_utils/server_floating_ip.py
+++ b/os_migrate/plugins/module_utils/server_floating_ip.py
@@ -63,20 +63,39 @@ class ServerFloatingIP(resource.Resource):
         raise exc.Unsupported("Direct ServerFloatingIP.create_or_update call is unsupported.")
 
     def create(self, conn, sdk_srv, mode):
-        mode_choices = ['auto', 'skip']
+        mode_choices = ['auto', 'skip', 'new', 'existing']
         if mode not in mode_choices:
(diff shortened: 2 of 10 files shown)


os_migrate/tests/unit/test_server_floating_ip.py
```diff
@@ -118,3 +119,49 @@ def test_server_floating_ip_sdk_params(self):
...
+        conn = mock.Mock()
+        conn.network.ips.return_value = [found_fip]
+
+        sdk_net = sdk_server_floating_ip()
+        fip = ServerFloatingIP.from_sdk(None, sdk_net)  # conn=None
+
+        self.assertEqual(found_fip, fip._find_specified_detached_floating_ip(conn))
+        conn.network.ips.assert_called_with(floating_ip_address='172.20.9.135')
+
+    def test_find_specified_detached_floating_ip_not_found(self):
```
jistr, line 147: Will add test case for 'not found but required'

Makefile
```diff
@@ -16,7 +16,7 @@ install: os_migrate-os_migrate-latest.tar.gz
 		source /root/venv/bin/activate; \
 	fi; \
 	cd releases; \
-	ansible-galaxy collection install --force os_migrate-os_migrate-latest.tar.gz
+	ansible-galaxy collection install $(OS_MIGRATE_INSTALL_ARGS) --force os_migrate-os_migrate-latest.tar.gz
```
jistr, line 19: Will split this out into its own commit. It allows passing `--no-deps` to the collection install, which speeds up commands like `./toolbox/run make test-unit` by a lot.

os_migrate/plugins/module_utils/server_floating_ip.py
```diff
@@ -63,21 +63,40 @@ def create_or_update(self, conn, filters=None):
         raise exc.Unsupported("Direct ServerFloatingIP.create_or_update call is unsupported.")
 
     def create(self, conn, sdk_srv, mode):
-        mode_choices = ['auto', 'skip']
+        mode_choices = ['auto', 'skip', 'new', 'existing']
```
jistr, line 66: Will add docs into the patch too

ccamacho (approved):

Makefile
philroche: Nice addition

PR #497, opened 2021-10-14
new: include an openstacksdk check in the preludes roles

ccamacho:
This commit adds a package version check in the prelude_src
and prelude_dst roles so we verify the user has a known working
openstacksdk version before running the migrations.

### DIFF
--- a/docs/src/roles/role-prelude_common.rst
+++ b/docs/src/roles/role-prelude_common.rst
@@ -1,0 +1,6 @@
+=====================
+Role - prelude_common
+=====================
+
+.. ansibleautoplugin::
+  :role: os_migrate/roles/prelude_common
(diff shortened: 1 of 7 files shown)


ccamacho:
Im testing a higher openstacksdk version to make it fail un purpose

ccamacho:
It seems that after every run the CI logs are pruned:

OK run:
```
TASK [os_migrate.os_migrate.prelude_src : fail if the user has an unsupported openstacksdk version] ***
Thursday 14 October 2021  13:12:48 +0000 (0:00:00.028)       0:01:09.604 ****** 
ok: [localhost] => changed=false 
  msg: |-
    The installed openstacksdk version 0.59.0
    is >= than 0.36.0"
```

FAIL run

PR #502, opened 2021-11-05
Remove filters from _refs_from_ser, create constant for %auth%

jistr:
(see commit messages)

### DIFF
--- a/os_migrate/tests/unit/test_resource.py
+++ b/os_migrate/tests/unit/test_resource.py
@@ -34,7 +34,7 @@ class FakeResource(resource.Resource):
     def _find_sdk_res(conn, name_or_id, filters=None):
         return valid_fakeresource_sdk()
 
-    def _refs_from_ser(self, conn, filters=None):
+    def _refs_from_ser(self, conn):
         return {
             'param3name': 'param3nameval',
             'param4name': 'param4nameval',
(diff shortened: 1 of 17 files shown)


os_migrate/tests/unit/test_resource.py
```diff
@@ -280,8 +280,9 @@ def test_dst_prerequisites_errors(self):
         res = FakeResource.from_data(valid_fakeresource_data())
         self.assertEqual(res.dst_prerequisites_errors(None), [])
 
-        def exception_refs_from_ser(_self, conn, filters=None):
+        def exception_refs_from_ser(_self, conn):
             raise ResourceFailure("Image 'asdf' not found.")
-        res._refs_from_ser = exception_refs_from_ser
+        # from a function to a bound method via __get__
+        res._refs_from_ser = exception_refs_from_ser.__get__(res, res.__class__)
```
jistr, line 286: Took me a bit to find what was wrong here... The function needs to be bound to the instance via __get__, shouldn't be just assigned as-is. Previously it worked by accident because of the last defaulted parameter, but the parameters were erroneously shifted by 1 when actually calling the method. Removing the filters parameter exposed this.

jistr:
Each time a different failure so far on 13-16, trying again.

philroche (approved):
Tests have passed and changes look good to me.

jistr:
Thanks :)

PR #507, opened 2021-11-30
Fix: enforce the executable argument when executing shell

ccamacho:
Currently the migrator host is not limited to be only
CentOS, Fedora, or RHEL. In the case a user runs OS
migrate from a Linux distribution which is not using
Bash as the default terminal, options like pipefail
might not be availabe (i.e. Ubuntu and ZSH) for
those cases we need to make explicit that the shell
we will use is Bash.

This commit enforce to use and set explicitly
the executable when running ansible.builtin.shell.

### DIFF
--- a/os_migrate/plugins/module_utils/resource.py
+++ b/os_migrate/plugins/module_utils/resource.py
@@ -300,7 +300,7 @@ class Resource():
         """
         Returns: True if resource data is well-formed, False otherwise
         """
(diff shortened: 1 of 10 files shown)


jistr:
What's the reasoning behind this? (We should add that info to the commit message too...)

jistr (approved):
Thanks!

PR #511, opened 2021-12-22
Docs: fix link on install-from-galaxy which address to install-from-source

jbadiapa:
The link on the user/install-from-galaxy.html address to user/install-from-source.md, which does not exit,  instead of user/install-from-source.html


ccamacho:
This is awesome!!! Thanks @jbadiapa for your contributions!!!! Let's wait for the CI to run and render the docs.

ccamacho (approved):

PR #514, opened 2022-01-07
Dev: Fix and extend e2e dev env setup instructions

jistr:
The `--id auto` parameter has to be removed from the flavor creation
because it tries to literally set flavor IDs to "auto", and as we're
creating 2 of those, they conflict. We should simply remove the `--id`
parameter to get the ID autogenerated.

The flavor RAM for the conversion host is increased so that it doesn't
get OOM killed.

We also add instructions on how to create public network, with
parameters for environments deployed with Infrared Virsh provisioning
plugin.

Minor formatting fixes are included.


ccamacho (approved):
docs-only: LGTM, the CI is blocked if we don't get Devstack fixed today at 12:00 I'll merge this by excluding the merge rules.

PR #516, opened 2022-02-03
Fix: Replace akward openstacksdk version check

noonedeadpunk:
Using shell for getting pip version leads to several highly possible
side effects. As example, you verify opensatcksdk persistance against
ENV however following tasks will use ansible_python_interpreter that
likely differ if virtualenv is used for ansible.

Another side effect is that openstacksdk could be installed in venv
which is not respected by shell when you provide interpreter for host
explicitly.

With this patch we replace shell with pip_package_info module that
collects all installed packages against ansible_python_interpeter.

### DIFF
--- a/os_migrate/roles/prelude_common/tasks/main.yml
+++ b/os_migrate/roles/prelude_common/tasks/main.yml
@@ -15,25 +15,24 @@
 # under the License.
 
 - name: get the current openstacksdk version
-  ansible.builtin.shell: |
-    set -eo pipefail
-    python3 -m pip show openstacksdk | grep Version | cut -d' ' -f 2
-  args:
-    executable: /bin/bash
+  community.general.pip_package_info:
   register: _installed_openstacksdk
-  changed_when: "_installed_openstacksdk.rc == 0"
+
+- name: set openstacksdk version
+  ansible.builtin.set_fact:
+    _installed_openstacksdk_versions: "{{ _installed_openstacksdk['packages']['pip']['openstacksdk'] | map(attribute='version') | list | first }}"
 
 - name: print the current installed version of openstacksdk
   ansible.builtin.debug:
-    var: _installed_openstacksdk.stdout
+    var: _installed_openstacksdk_versions
 
 - name: fail if the user has an unsupported openstacksdk version
   ansible.builtin.assert:
     that:
-      - _installed_openstacksdk.stdout is version('{{ os_migrate_common_minimum_openstacksdk_version|string }}', '>=')
+      - _installed_openstacksdk_versions is version(os_migrate_common_minimum_openstacksdk_version, '>=')
     fail_msg: |
-      The installed openstacksdk version {{ _installed_openstacksdk.stdout }}
+      The installed openstacksdk version {{ _installed_openstacksdk_versions }}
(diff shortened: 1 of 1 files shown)


os_migrate/roles/prelude_common/tasks/main.yml
```diff
@@ -15,25 +15,24 @@
...
-    set -eo pipefail
-    python3 -m pip show openstacksdk | grep Version | cut -d' ' -f 2
-  args:
-    executable: /bin/bash
+  community.general.pip_package_info:
   register: _installed_openstacksdk
-  changed_when: "_installed_openstacksdk.rc == 0"
+
+- name: set openstacksdk version
+  set_fact:
```
ccamacho, line 22: This needs to be ansible.builtin.set_fact

ccamacho:
Thanks @noonedeadpunk for your contributions!!! Just a simple change to pass the lint checks, `set_fact` must be `ansible.builtin.set_fact`.

Also if you could change the commit title to `Fix: Replace akward openstacksdk version check` we can show it correctly in the changelog.

Thanks!!!

os_migrate/roles/prelude_common/tasks/main.yml
```diff
@@ -15,25 +15,24 @@
...
+    _installed_openstacksdk_versions: "{{ _installed_openstacksdk['packages']['pip']['openstacksdk'] | map(attribute='version') | list }}"
 
 - name: print the current installed version of openstacksdk
   ansible.builtin.debug:
-    var: _installed_openstacksdk.stdout
+    var: _installed_openstacksdk_versions[0]
```
odyssey4me, line 27: ``_installed_openstacksdk_versions | first`` might be considered slightly more elegant... but meh

os_migrate/roles/prelude_common/tasks/main.yml
```diff
@@ -15,25 +15,24 @@
...
-    python3 -m pip show openstacksdk | grep Version | cut -d' ' -f 2
-  args:
-    executable: /bin/bash
+  community.general.pip_package_info:
   register: _installed_openstacksdk
-  changed_when: "_installed_openstacksdk.rc == 0"
+
+- name: set openstacksdk version
+  set_fact:
+    _installed_openstacksdk_versions: "{{ _installed_openstacksdk['packages']['pip']['openstacksdk'] | map(attribute='version') | list }}"
```
odyssey4me, line 23: could add ``| first`` at the end of this so that we only get a single entry... seeing that's all we use

noonedeadpunk: Yep, great idea to use first. Just `[0]` would be so messy here :p

os_migrate/roles/prelude_common/tasks/main.yml
noonedeadpunk: Done

jistr (approved):
Thanks for the patch, looks good! Triggering e2e CI.

jistr:
Sorry, we have issues with the CI setup, it may take some time to solve.

ccamacho (approved):
Thank you!!!

PR #517, opened 2022-02-03
New: Attaching conversion hosts onto public networks

jistr:
This PR combines 3 new features to solve a new RFE: attaching conversion hosts directly onto public networks. Apart from documentation for the small individual features, the variables guide also specifically addresses the use case of conversion hosts on public networks.


jistr:
On review for a while, with green CI and we are short on active reviewers now, so i'll self-merge.

PR #527, opened 2022-03-31
dev: update ansible version in toolbox container

cloudguruab:
```
Error Summary
-------------
Disk Requirements:
   At least 287MB more space needed on the / filesystem.

Error: error building at STEP "RUN bash /build/build.sh": error while running runtime: exit status 1
make: *** [Makefile:176: toolbox-build] Error 125
```
- Ran out of memory for local test. What is next best step to solve this? I recall you mentioning this being an issue before for our vm's


jistr:
Does `make test-fast` work for you with the change? I'm getting:

```
ERROR: Found 74 validate-modules issue(s) which need to be resolved:                                                                                                                                               
ERROR: plugins/modules/auth_info.py:0:0: missing-gplv3-license: GPLv3 license header not found in the first 20 lines of the module                           
...
...
```

We probably need to add an ignore-2.12.txt file here:

https://github.com/os-migrate/os-migrate/tree/main/os_migrate/tests/sanity

(just a copy of 2.11) and see if it works.

jistr:
I just tried adding that file and there are still multiple errors. Changing the Ansible version seems to be more breaking than i expected it to be, and given the way it's implemented, CI cannot tell us if anything's wrong. I think we'll have to approach this from a different end, perhaps start with a different PR and then get back to this one.

We'd start with consolidating the CI job matrix. I'd change the matrix [1] from `[2.9, 3.4, 4.5]` to `[2.9, latest]` and change the venv-requirements files [2] accordingly. We would have `venv-requirements-ansible-2.9.txt` and `venv-requirements-ansible-latest.txt`. For now, i'd keep Ansible 4.5 in `...latest.txt`. That gives us CI running on 2.9 and 4.5, and then we can try moving the 4.5 in `latest` to 5.X and deal with the errors. The important thing about this approach is that if 5.X brings any errors, CI will tell us about those (currently CI is green even though the move to 5.X is breaking the tests locally).

[1] https://github.com/os-migrate/os-migrate/blob/main/.github/workflows/consistency-functional.yml#L26
[2] https://github.com/os-migrate/os-migrate/tree/main/toolbox/build

cloudguruab:
@jistr adding `ignore-2.12.txt` file had no overall impact, but it did bring the validate-modules count down to 31

jistr (approved):

jistr:
Continued networking issues in CI infrastructure. The patch is primarily affecting the 13-16 job which uses the toolbox Ansible version. 10-16 is pinned to Ansible 2.9 and shouldn't be affected by the patch. Merging.

PR #530, opened 2022-04-04
dev: consolidate ci matrix for ansible latest

cloudguruab:
### Notes from #527 w/ Jistr

I just tried adding that file and there are still multiple errors. Changing the Ansible version seems to be more breaking than i expected it to be, and given the way it's implemented, CI cannot tell us if anything's wrong. I think we'll have to approach this from a different end, perhaps start with a different PR and then get back to this one.

We'd start with consolidating the CI job matrix. I'd change the matrix [1] from [2.9, 3.4, 4.5] to [2.9, latest] and change the venv-requirements files [2] accordingly. We would have venv-requirements-ansible-2.9.txt and venv-requirements-ansible-latest.txt. For now, i'd keep Ansible 4.5 in ...latest.txt. That gives us CI running on 2.9 and 4.5, and then we can try moving the 4.5 in latest to 5.X and deal with the errors. The important thing about this approach is that if 5.X brings any errors, CI will tell us about those (currently CI is green even though the move to 5.X is breaking the tests locally).


cloudguruab:
Ran into this error locally, also want to check that I am following this correctly: 

```
Traceback (most recent call last):
  File "/root/venv/bin/ansible-test", line 28, in <module>
    main()
  File "/root/venv/bin/ansible-test", line 22, in main
    from ansible_test._internal.cli import main as cli_main
ImportError: cannot import name 'main' from 'ansible_test._internal.cli' (/root/venv/lib64/python3.8/site-packages/ansible_test/_internal/cli/__init__.py)
make: *** [Makefile:133: test-sanity] Error 1
```

.github/workflows/consistency-functional.yml
```diff
@@ -23,7 +23,7 @@ jobs:
       max-parallel: 4
       matrix:
         python-version: [3.7]
-        ansible-version: [2.9, 3.4, 4.5]
+        ansible-version: [2.9, 4.5]
```
jistr, line 26: In spirit yes, let's also rename '4.5' to 'latest' so that we don't have to keep renaming files and editing many places whenever we bump the Ansible there to something newer?

jistr: We'll also want to make the same matrix change for the functional test below on line 101.

jistr (changes requested):

jistr:
Looks good, we should just squash the commits so that the PR has just one 'dev' commit. (`git rebase --interactive` - i can help on tomorrow's sync)

cloudguruab:
Sounds good, also passing locally as well!

jistr (approved):

PR #537, opened 2022-04-21
fix: Export of subnets with multiple allocation pools

jistr:
Subnets with multiple allocation pools previously crashed on export
due to an issue with parameter sorting. This is now fixed.


cloudguruab (approved):
LGTM!

PR #541, opened 2022-05-16
dev: potential patch for continued work in b/test-ci

cloudguruab:


cloudguruab:
https://github.com/pypa/pip/issues/10373#issuecomment-902551157

PR #543, opened 2022-05-26
F/import role assignment

cloudguruab:


cloudguruab:
Running into some weird error locally outside of simple lint fixes. Might just be some small syntax issue.

jistr:
I'm not sure why that method isn't available, but perhaps we could solve it by using the `role_assignments(**query)` method instead -- pass all 3 IDs to limit the role assignments we're looking for. If we get a list with 1 item, the assignment exists, if we get a list with 0 items, it doesn't.

https://docs.openstack.org/openstacksdk/latest/user/proxies/identity_v3.html#role-assignment-operations

os_migrate/plugins/module_utils/const.py
```diff
@@ -11,7 +11,7 @@
 RES_TYPE_IMAGE = 'openstack.image.Image'
 RES_TYPE_KEYPAIR = 'openstack.compute.Keypair'
 RES_TYPE_NETWORK = 'openstack.network.Network'
-RES_TYPE_USER_PROJECT_ROLE_ASSIGNMENT = 'openstack.identity.UserProjectRoleAssignment'
+RES_TYPE_USER_PROJECT_ROLE_ASSIGNMENT = 'openstack.identity.RoleAssignment'
```
jistr, line 14: I think this should stay "UserProjectRoleAssignment" as it was before, we're specifying the resource type as seen by OS Migrate. Usually it maps to OpenStack resources 1:1 but it does not have to, e.g. the OS Migrate "RouterInterface" type maps to OpenStack's "Port" type. Every router interface is a port, but not every port is a router interface.

Similarly here - every UserProjectRoleAssignment is an OpenStack role assignment, but not every OpenStack role assignment is between a user-project-role triplet, which is what we're handling here.

This rename can be done in a separate mini-patch. The rest of the PR looks good, thanks for implementing this!

jistr (approved):

PR #544, opened 2022-06-02
dev: Pin test/lint dependencies to reduce random CI issues

jistr:
We are getting random breakages whenever something changes in the best
practices of some lint dependencies. Let's pin the test+lint
dependencies so that we tackle the updating of these willingly, rather
than randomly being hit with CI breakage.


cloudguruab (approved):
LGTM

PR #545, opened 2022-06-10
new: skip serialization if dns_domain is false

cloudguruab:
- Tested in unit and passes, ended up commenting out - not sure if this would even require its own testcase of any sorts?

### DIFF
--- a/os_migrate/tests/unit/test_network.py
+++ b/os_migrate/tests/unit/test_network.py
@@ -162,3 +162,12 @@ class TestNetwork(unittest.TestCase):
         # disallowed params when creating a network
         self.assertNotIn('availability_zones', sdk_params)
         self.assertNotIn('revision_number', sdk_params)
+
+    def test_network_import_when_dns_domain_is_false(self):
+        ser = serialized_network()
+        ser['params']['dns_domain'] = ''
+        net = Network.from_data(ser)
+        refs = net._refs_from_ser(None)
+        sdk_params = net._to_sdk_params(refs)
+
+        self.assertNotIn('dns_domain', sdk_params)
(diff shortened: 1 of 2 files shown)


cloudguruab:
@jistr going to fail, but trying to implement a mock of serialized object params and sdk_params to tweak by emptying the dns_domain key and then checking its presence in both params, and sdk_params

os_migrate/tests/unit/test_network.py
```diff
@@ -162,3 +163,24 @@ def test_network_sdk_params(self):
...
+    def test_network_serialization_when_dns_domain_is_false(self):
+        sdk_net = sdk_network()
+        net = Network.from_sdk(None, sdk_net)  # conn=None
+        params, info = net.params_and_info()
+        
+        with mock.patch.dict(params['dns_domain'], {'dns_domain': ''}):
+            self.assertEqual(params['dns_domain'], 'example.org')
+
+    def test_network_import_when_dns_domain_is_false(self):
+        net = Network.from_data(serialized_network())
```
jistr, line 181: I think we need to save the serialized_network() result to a variable, which will allow us to set the dns_domain to empty string, and only then we feed it to Network.from_data, and we can then validate that it is not present in sdk_params.

cloudguruab: Ahhhh looks like I was over complicating! - Thanks

jistr:
Nice, let's just squash into a single `new` or `fix` commit (this is a user-facing change so it should appear in changelog) and we can merge. :+1:

cloudguruab:
@jistr will merge documentation #546 first then rebase this branch ontop of that & merge

jistr (approved):

PR #546, opened 2022-06-14
dev: documentation update - no_log issue, import role assign patch

cloudguruab:


cloudguruab:
- will require rebase

docs/src/modules/module-import_user_project_role_assignment.rst
```diff
@@ -1,6 +1,6 @@
-============================================
-Module - import_exportuser_project_role_assignment
-============================================
+===================================================
+Module - import_export_user_project_role_assignment
```
jistr, line 2: It says 'import_export_...', it should be just 'import_...'

The rest looks good :+1: . Next time i think we should have separate pull requests for separate concerns. The dns_domain being one, and the no_log issue being another. The typo fix here could be piggy-backed onto any of those PRs (it's very small and simple change) or a separate PR as well. But here i think let's keep it as-is in a single PR.

Bonus points if we get to rebase/squash things into 3 separate commits. The documentation commits should probably be `dev`, while the dns_domain commit should be `new` or `fix`.

jistr (approved):

PR #549, opened 2022-06-30
new: writing out clouds.yml file

cloudguruab:


cloudguruab:
@ccamacho I dont have the e2e env setup on my lab machine so not sure if this will lead to issues running/testing this - this is what I have for writing out the clouds.yml file. Let me know if anything should be changed.

PR #551, opened 2022-07-12
new: write clouds.yaml file

cloudguruab:


jistr:
It would be nice to post in a comment how the resulting clouds.yaml file looks in functional test execution. I wonder if the indentation will be correct for YAML formatting, but it's something hard to nail exactly without some amount of trial and error, so it's also hard to address in a review without seeing the generated file.

jistr:
But the patch looks good from overall direction point of view :)

cloudguruab:
Accidentally destroyed this PR locally

PR #552, opened 2022-07-14
new: Write auth config to clouds.yaml

cloudguruab:
Yaml format for clouds file is a little weird - will play with indents. 

```
clouds:
  src:
      auth: auth_url: http://192.168.122.4/identity
  password: password
  project_domain_id: default
  project_name: demo
  user_domain_id: default
  username: demo

          region_name: RegionOne

    dst:
      auth: auth_url: http://192.168.122.4/identity
  password: password
  project_domain_id: default
  project_name: alt_demo
  user_domain_id: default
  username: alt_demo_member

          region_name: RegionOne
```

### DIFF
--- a/os_migrate/roles/prelude_common/templates/clouds.yaml.j2
+++ b/os_migrate/roles/prelude_common/templates/clouds.yaml.j2
@@ -1,0 +1,23 @@
+clouds:
+  src:
+{% if os_migrate_src_auth is defined and os_migrate_src_auth|length %}
+    auth: 
+      {{os_migrate_src_auth|to_nice_yaml(indent=2)|indent(6)}}
+{%- endif %}
+{% if os_migrate_src_auth_type is defined %}
+    auth_type: {{os_migrate_src_auth_type}}
+{%- endif %}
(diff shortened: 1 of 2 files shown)


os_migrate/roles/prelude_common/templates/clouds.yaml.j2
```diff
@@ -0,0 +1,21 @@
+clouds:
+  src:
+  {% if os_migrate_src_auth is defined and os_migrate_src_auth|length %}
+    auth: {{os_migrate_src_auth|to_nice_yaml(indent=2)|indent(2)}}
```
jistr, line 4: I think this one will have to be on 2 lines to format it like a proper YAML dict, somehting like:

```
auth:
  {{os_migrate_src_auth|to_nice_yaml(indent=2)|indent(2)}}
```

But the indentation numbers are probably off still.

jistr (approved):
Looks good :+1: . We can go further and make the path configurable and perhaps default it to `auth` subdirectory of the data dir, rather than put it directly into data dir among the resource YAMLs, but i think that can be refined in a separate patch.

PR #553, opened 2022-07-15
new: Sort exported server ports as they appear in Nova API

jistr:
We export server port info from Neutron API, because it gives more
information about the port. However, querying ports from Neutron seems
to give a somewhat random order of the ports.

Now we still export the same Neutron info about ports, but we sort
them based on the order in which the VM IP addresses appear in
OpenStack SDK's 'compute.get_server' return value.

This approach cannot guarantee that the original order of NICs is
preserved during migration (Nova does not guarantee NIC ordering [1]),
but it makes it more likely that the NIC order will be preserved
during migration.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1798126

### DIFF
--- a/os_migrate/plugins/module_utils/server_port.py
+++ b/os_migrate/plugins/module_utils/server_port.py
@@ -6,11 +6,46 @@ import openstack
 from ansible_collections.os_migrate.os_migrate.plugins.module_utils \
     import exc, const, reference, resource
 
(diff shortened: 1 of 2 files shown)


cloudguruab (approved):
LGTM!

ccamacho (approved):

PR #554, opened 2022-07-21
chg: update to network role/play for auth replacement

cloudguruab:
- Let me know how this looks: 
- Also running into an error in prelude_common role locally which is weird, something with setting openstack sdk version


os_migrate/playbooks/export_networks.yml
```diff
@@ -1,3 +1,6 @@
 - hosts: migrator
+  environment:
+    OS_CLIENT_CONFIG_FILE: "{{ os_migrate_data_dir }}/clouds.yaml"
+    OS_CLOUD: "dst"
```
jistr, line 4: We shouldn't specify OS_CLOUD here, as they may be playbooks that use both src and dst. Rather we should hard-code `cloud: src` wherever we used `os_migrate_src_auth` previously, and respectively for `dst`.

os_migrate/playbooks/export_networks.yml
```diff
@@ -1,3 +1,6 @@
 - hosts: migrator
+  environment:
+    OS_CLIENT_CONFIG_FILE: "{{ os_migrate_data_dir }}/clouds.yaml"
```
jistr, line 3: This should work, but as we talked before about configurability of this, we may consider adding a `os_migrate_clouds_path` variable and using the above as default. So it would become:

```
OS_CLIENT_CONFIG_FILE: "{{ os_migrate_clouds_path|default(os_migrate_data_dir ~ '/clouds.yaml') }}"
```

os_migrate/roles/prelude_common/tasks/main.yml
```diff
@@ -36,3 +36,8 @@
     success_msg: |
       The installed openstacksdk version {{ _installed_openstacksdk_versions }}
       is >= than {{ os_migrate_common_minimum_openstacksdk_version }}
+
+- name: write to auth file
+  ansible.builtin.template:
+    src: clouds.yaml.j2
+    dest: "{{ os_migrate_data_dir }}/clouds.yaml"
```
jistr, line 43: And we'd edit this to use the same `"{{ os_migrate_clouds_path|default(os_migrate_data_dir ~ '/clouds.yaml') }}"`

os_migrate/roles/prelude_common/templates/clouds.yaml.j2
```diff
@@ -0,0 +1,23 @@
+clouds:
```
jistr, line 1: I think these parts were already merged, there may need to be a rebase of the PR to make this code disappear from the PR.

os_migrate/playbooks/export_networks.yml
jistr: We specify OS_CLIENT_CONFIG_FILE which points the OpenStack SDK at the file which defines src and dst variables, and then each module call will have hard-coded either `cloud: src` or `cloud: dst` specifying which entry in the file should be used. IIUC, the OS_CLIENT_CONFIG_FILE variable would be global for the whole Ansible play, not just a role.

os_migrate/roles/prelude_common/templates/clouds.yaml.j2
jistr: All the auth replacement PRs should be merging into `feature/auth-replacement`. I don't recall what we needed to change in Makefile tho :)

jistr:
The total PR diff is looking good, we'll need commit squashing to keep the changelog clean, and we'll need to figure out the test failure. The error message doesn't point to anything that would be obvious to me though... @ccamacho may be able to help with figuring it out.

PR #557, opened 2022-08-01
new: cloud config for network resources

cloudguruab:
Running the following functional test locally:
```
 OS_MIGRATE_FUNC_TEST_ARGS='--tags test_network' ./toolbox/run make test-func
```

Response: 
```
failed: [localhost] (item={'auth_url': 'http://192.168.122.4/identity', 'password': 'password', 'project_domain_id': 'default', 'project_name': 'demo', 'user_domain_id': 'default', 'username': 'demo'}) => changed=false 
  ansible_loop_var: item
  extra_data:
    data: null
    details: 'Unable to complete operation on network 0cfb4319-83c3-43ab-9512-9eefb0eda9b1. There are one or more ports still in use on the network, id for these ports is: c31847b5-6d52-4454-ab11-839d045c6fa1,d5375312-138f-4267-ab91-dbb0051c508f.'
    response: '{"NeutronError": {"type": "NetworkInUse", "message": "Unable to complete operation on network 0cfb4319-83c3-43ab-9512-9eefb0eda9b1. There are one or more ports still in use on the network, id for these ports is: c31847b5-6d52-4454-ab11-839d045c6fa1,d5375312-138f-4267-ab91-dbb0051c508f.", "detail": ""}}'
  item:
    auth_url: http://192.168.122.4/identity
    password: password
    project_domain_id: default
    project_name: demo
    user_domain_id: default
    username: demo
  msg: 'ConflictException: 409: Client Error for url: http://192.168.122.4:9696/v2.0/networks/0cfb4319-83c3-43ab-9512-9eefb0eda9b1, Unable to complete operation on network 0cfb4319-83c3-43ab-9512-9eefb0eda9b1. There are one or more ports still in use on the network, id for these ports is: c31847b5-6d52-4454-ab11-839d045c6fa1,d5375312-138f-4267-ab91-dbb0051c508f.'
  ```

### DIFF
--- a/os_migrate/playbooks/export_flavors.yml
+++ b/os_migrate/playbooks/export_flavors.yml
@@ -1,5 +1,3 @@
 - hosts: migrator
-  environment:
-    OS_CLIENT_CONFIG_FILE: "{{ os_migrate_clouds_path|default(os_migrate_data_dir ~ '/clouds.yaml') }}"
   roles:
     - os_migrate.os_migrate.export_flavors
--- a/os_migrate/playbooks/export_images.yml
+++ b/os_migrate/playbooks/export_images.yml
@@ -1,5 +1,3 @@
 - hosts: migrator
-  environment:
-    OS_CLIENT_CONFIG_FILE: "{{ os_migrate_clouds_path|default(os_migrate_data_dir ~ '/clouds.yaml') }}"
   roles:
     - os_migrate.os_migrate.export_images
--- a/os_migrate/playbooks/export_keypairs.yml
+++ b/os_migrate/playbooks/export_keypairs.yml
@@ -1,5 +1,3 @@
 - hosts: migrator
-  environment:
-    OS_CLIENT_CONFIG_FILE: "{{ os_migrate_clouds_path|default(os_migrate_data_dir ~ '/clouds.yaml') }}"
   roles:
     - os_migrate.os_migrate.export_keypairs
--- a/os_migrate/playbooks/export_networks.yml
+++ b/os_migrate/playbooks/export_networks.yml
@@ -1,6 +1,5 @@
 - hosts: migrator
-  vars:
-    env:
-      OS_CLIENT_CONFIG_FILE: "{{ os_migrate_clouds_path|default(os_migrate_data_dir ~ '/clouds.yaml') }}"
+  environment:
+    OS_CLIENT_CONFIG_FILE: "{{ os_migrate_clouds_path|default(os_migrate_data_dir ~ '/clouds.yaml') }}"
   roles:
     - os_migrate.os_migrate.export_networks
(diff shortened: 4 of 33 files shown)

### CODE REVIEW

**Comment by jistr** [comment]
As i debugged this i had to make code changes to understand where the problem is, so rather than describe everything in review, i just pushed the code changes - with this, when i ran:

```
OS_MIGRATE_INSTALL_ARGS='--no-deps' OS_MIGRATE_FUNC_TEST_ARGS='--tags test_network' ./toolbox/run make test-func
```

... the func tests on network export/import pass, and it's using clouds.yaml 🎉.

The main issue was that in the func tests we're actually not importing the full `playbooks/export_networks.yml`, so whatever we do there has no effect in the func test. Ansible doesn't allow pulling in playbooks inside roles, so we have to pull in individual roles inestead, like we do here: https://github.com/os-migrate/os-migrate/blob/main/tests/func/tasks/tenant/run/network.yml#L28-L30 .

That means the `OS_CLIENT_CONFIG_FILE` setting has to be done in the same way but in the test playbook to make it work in tests too. If we can work our way through the other resources in the spirit of this commit https://github.com/os-migrate/os-migrate/pull/557/commits/3d41a476e363cadbb7b75af86c677052f704ef25 then i think we'd get the func tests going.

There may be other similar changes that we later need to do to e2e playbooks, but i wouldn't worry about it until we have the func tests green.

One suggestion - if we want to chunk it up into smaller bits to make it easier to work on, i'd suggest still focusing on a single resource for now. If we do sweeping changes to many files and then in review/testing we discover that we want to amend something, we'll have to make edits in many places. It might be easier getting just one resource (networks) fully replaced first, get all the tests passing incl. e2e, merge the PR and then move onto other resources. The other resources could then perhaps be done all in a single PR, since we'd already be following a set convention and not doing solution discovery, so i think you could save all the work on the other resources you've done so far for that 2nd PR.

The general pattern "get the approach working and polished on one piece of code, then do the same refactoring to the rest of the code base" can save a lot of work when doing big refactorings like this. It's just a suggestion though. If you feel more comfortable editing all at once and don't mind potentially making amendments in many places, that's fine too.

**Comment by jistr** [comment]
The Github interface doesn't make this too apparent so i'll mention it explicitly - i pushed one new commit into the PR itself (which updates the PR branch in your repo fork). So you may want to fetch/pull from your repo fork branch into your local branch to get the changes.

All in all this is on a good track and it was like that all along, just the environment setting in the individual playbooks not taking effect in tests was the gotcha.

**Comment by jistr** [comment]
And also small changes to the python modules are necessary (as visible from the commit i pushed), but that only became apparent once the `OS_CLIENT_CONFIG_FILE` environment var took effect.

**Comment by cloudguruab** [comment]
Getting the following error running this command ` openstack image create --public --disk-format raw --file cirros-0.4.0-x86_64-disk.img cirros-0.4.0-x86_64-disk.img`
```
HttpException: 502: Server Error for url: http://192.168.122.4/image/v2/images/6e49a0c4-265c-41df-9a54-782ede9998c8/file, The proxy server received an invalid: 502 Bad Gateway: response from an upstream server.: Bad Gateway
```

PR #562, opened 2022-08-19
new: flavor resource auth replacement

cloudguruab:


jistr:
Nice rebases the PRs. I looked at the logs, the tests are failing because the job is using old image with openstack.cloud collection dependency on version 1.8.0 (still with 'openstack_image'), while the test code had a patch for the 1.9.0 (with 'image'). I re-ran test image build+push job at https://github.com/os-migrate/os-migrate/actions/runs/2919280982 and once it finishes, we can try re-checking the PRs.

jistr (approved):

PR #563, opened 2022-08-19
new: projects resource auth replacement

cloudguruab:


jistr (changes requested):
The PR should be submitted against the feature/auth-replacement branch.

PR #577, opened 2022-11-17
validate changelog 

cloudguruab:
Running lints on commit messages is merged into `Makefile` as `lint-commit-messages`. Updated our github ci as well to include a test for validating changelog.


cloudguruab:
Expecting this to fail due to first commit message breaking rules, will rebase and consolidate messages after case is met.

PR #578, opened 2022-11-28
new: changelog validation

cloudguruab:
- Running into a range error for git log due to force push, https://stackoverflow.com/questions/19227219/dealing-with-git-fatal-invalid-revision-range


jistr:
Perfect :+1:

jistr (approved):

PR #588, opened 2023-01-17
dev: Fix functional CI

jistr:
This is an attempt to fix the following error in CI:

The following packages have unmet dependencies:
 postgresql-server-dev-all : Depends: postgresql-common (= 238) but 246.pgdg22.04+1 is to be installed
E: Unable to correct problems, you have held broken packages.

### DIFF
--- a/.github/workflows/consistency-functional.yml
+++ b/.github/workflows/consistency-functional.yml
@@ -142,6 +142,7 @@ jobs:
           sudo locale-gen en_US.UTF-8
           sudo dpkg-reconfigure locales
           sudo apt update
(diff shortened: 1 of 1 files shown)


cloudguruab:
Debugging in another branch but came across this: https://launchpad.net/ubuntu/+source/postgresql-common/238, looks like the changelog for the package version we are dependent on says `unstable`. Assuming we're running the latest version of ubuntu in our ci we could revert to using a different postgres package: https://packages.ubuntu.com/source/bionic/postgresql-common

jistr:
Neat, i'll close this PR then.

PR #594, opened 2023-03-13
dev: working on neutron ports

cloudguruab:


cloudguruab:
Updated the existing implementation but wondering if I should pursue creating a mock to test this.

os_migrate/plugins/module_utils/server_port.py
```diff
@@ -62,7 +62,10 @@ class ServerPort(resource.Resource):
         'fixed_ips_refs',
         'network_ref',
     ]
-    sdk_params_from_params = []
+    sdk_params_from_params = [
+        'id',
+        'binding_profile'
+    ]
```
jistr, line 68: I'd try with
```
params_from_sdk = [
        'binding_profile'
]
```

We don't need `id` (port id) here. We only need `network_id` and that is already being fed in because it's listed in `sdk_params_from_refs`.

cloudguruab:
`ConflictException`: Unable to complete operation on subnet 3b86636a-df29-48db-9c1b-f099f35abba9: One or more ports have an IP allocation from this subnet.'

Debugging this atm

os_migrate/plugins/module_utils/server_port.py
```diff
@@ -78,7 +80,12 @@ def from_sdk(cls, conn, sdk_resource):
         return obj
 
     def create_or_update(self, conn, filters=None):
-        raise exc.Unsupported("Direct ServerPort.create_or_update call is unsupported.")
+        refs = self._refs_from_ser(conn)
+        sdk_params = self._to_sdk_params(refs)
+
+        # creation using neutron, NOTE: add update method
+        sdk_port = conn.network.create_port(**sdk_params)
+        return {'port_id': sdk_port['id']}
```
jistr, line 88: The 'port_id' could be the culprit, see the docs re the 'networks' param for create_server: 

https://docs.openstack.org/api-ref/compute/?expanded=create-server-detail#create-server

PR #595, opened 2023-03-19
dev: working on failing testcase for role assignment

cloudguruab:
This PR is a proposed patch to [OSP-22603](https://issues.redhat.com/browse/OSP-22603). The bug is found on import for the user project role assignment module. `the result of import_user_project_role_assignment was "ok" even when the target user, project, and role exist but were not assigned.` This patch will contain a try/except for checking assignment of user project roles on import `create_or_update`, which will include an updated functional test case and unit test.


cloudguruab:
PR impacted by ci testing with port creation in #594, once functional tests pass - should be good to go

jistr (approved):

PR #600, opened 2023-04-05
new: port binding & creation via neutron api

cloudguruab:
This PR aims to cover https://issues.redhat.com/browse/OSP-18661 and will include a fresh deployment to ansible galaxy.

### DIFF
--- a/os_migrate/plugins/module_utils/server_port.py
+++ b/os_migrate/plugins/module_utils/server_port.py
@@ -85,7 +85,11 @@ class ServerPort(resource.Resource):
 
         # creation using neutron, NOTE: add update method
         sdk_port = conn.network.create_port(**sdk_params)
-        return {'id': sdk_port['id']}
+        return {'networks': [
+            {
+                'port': sdk_port['id']
+            }
+        ]}
 
     def nova_sdk_params(self, conn):
         refs = self._refs_from_ser(conn)
--- a/os_migrate/plugins/module_utils/server_port.py
+++ b/os_migrate/plugins/module_utils/server_port.py
@@ -85,7 +85,7 @@ class ServerPort(resource.Resource):
 
         # creation using neutron, NOTE: add update method
         sdk_port = conn.network.create_port(**sdk_params)
-        return {'port_id': sdk_port['id']}
+        return {'id': sdk_port['id']}
 
     def nova_sdk_params(self, conn):
         refs = self._refs_from_ser(conn)
--- a/os_migrate/plugins/module_utils/server_port.py
+++ b/os_migrate/plugins/module_utils/server_port.py
@@ -85,11 +85,9 @@ class ServerPort(resource.Resource):
 
         # creation using neutron, NOTE: add update method
         sdk_port = conn.network.create_port(**sdk_params)
-        return {'networks': [
-            {
-                'port': sdk_port['id']
-            }
-        ]}
+        return {
+            'port': sdk_port['id']
+        }
 
     def nova_sdk_params(self, conn):
         refs = self._refs_from_ser(conn)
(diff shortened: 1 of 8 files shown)


os_migrate/plugins/module_utils/server_port.py
```diff
@@ -78,7 +80,12 @@ def from_sdk(cls, conn, sdk_resource):
         return obj
 
     def create_or_update(self, conn, filters=None):
-        raise exc.Unsupported("Direct ServerPort.create_or_update call is unsupported.")
+        refs = self._refs_from_ser(conn)
+        sdk_params = self._to_sdk_params(refs)
+
+        # creation using neutron, NOTE: add update method
+        sdk_port = conn.network.create_port(**sdk_params)
+        return {'id': sdk_port['id']}
```
jistr, line 88: Why 'id'? See the doc on the 'networks' parameter how to pass in a port reference: https://docs.openstack.org/api-ref/compute/?expanded=create-server-detail#create-server 

Perhaps we should just return `sdk_port` from here and do the parameter molding entirely in the `update_sdk_params_networks_simple`, that would be cleaner and less confusing approach perhaps. It would make it more visible how the `networks` parameter is being formed.

jistr: networks | body | array | A list of network object. Required parameter when there are multiple networks defined for the tenant. When you do not specify the networks parameter, the server attaches to the only network created for the current tenant.  Optionally, you can create one or more NICs on the server. To provision the server instance with a NIC for a network, specify the UUID of the network in the uuid attribute in a networks object.  To provision the server instance with a NIC for an already existing port, specify the port-id in the port attribute in a networks object. ...
-- | -- | -- | --

...



networks.port (Optional) | body | string | To provision the server instance with a NIC for an already existing port, specify the port-id in the port attribute in a networks object. The port status must be DOWN. Required if you omit the uuid attribute. Requested security groups are not applied to pre-existing ports.
-- | -- | -- | --

jistr:
It's trying to make a bad CLI call because it uses the table border `+-----+` instead of the ID of the port from port list. We may need one more `--format value` if port list supports it. See here:

```
TASK [Check for ports with the existing subnet id] *****************************
changed: [localhost] => changed=true 
  cmd: |-
    openstack port list --fixed-ip subnet=1cd6bf00-a848-4c48-8b10-6d99cfc81614 --column ID
  delta: '0:00:01.493234'
  end: '2023-04-07 18:51:38.677276'
  msg: ''
  rc: 0
  start: '2023-04-07 18:51:37.184042'
  stderr: ''
  stderr_lines: <omitted>
  stdout: |-
    +--------------------------------------+
    | ID                                   |
    +--------------------------------------+
    | 8de42bf9-afca-4fe4-bd4f-f17a7a778dca |
    +--------------------------------------+
  stdout_lines: <omitted>
TASK [Check for ports with no device owner to be deleted] **********************
failed: [localhost] (item=+--------------------------------------+) => changed=false 
  ansible_loop_var: item
  cmd: |-
    openstack port show --format value --column device_owner +--------------------------------------+
  delta: '0:00:01.391252'
  end: '2023-04-07 18:51:40.298810'
  item: +--------------------------------------+
  msg: non-zero return code
  rc: 1
  start: '2023-04-07 18:51:38.907558'
  stderr: No Port found for +--------------------------------------+
  stderr_lines: <omitted>
  stdout: ''
  stdout_lines: <omitted>
failed: [localhost] (item=| ID                                   |) => changed=false 
```

os_migrate/plugins/module_utils/server_port.py
```diff
@@ -62,7 +62,9 @@ class ServerPort(resource.Resource):
         'fixed_ips_refs',
         'network_ref',
     ]
-    sdk_params_from_params = []
+    params_from_sdk = [
+        'binding_profile'
```
jistr, line 66: In a follow up PR we should add mac address here, that was one of the use cases i think.

PR #612, opened 2023-05-16
dev: working on initial changes for stg mig w/o CH

cloudguruab:
Initial changes for the storage migrations pieces. Currently working on booting from local boot disk. I have attached questions comments to code using `TODO-storage`, overall the following files contain questions associated with some changes: 

- [`conversion_host_inventory.yml`](https://github.com/os-migrate/os-migrate/pull/612/files#diff-a04a50cbf02d5db9cbf64fe17558cf67b75c2c1a17eb91ffe4f7a9e6309679caR30)
- [`export_workload.py`](https://github.com/os-migrate/os-migrate/pull/612/files#diff-f5b2ab8315a90d71b638ce4a14f20ea91f84ab2aa21aaf7ecd67552c083ddc64R125)
- [`server.py`](https://github.com/os-migrate/os-migrate/pull/612/files#diff-bfe9cb91daea9554c5afa52905e08effdd35295b42e340cc4d10daaf51b62e77R235)


os_migrate/plugins/module_utils/server.py
```diff
@@ -67,17 +67,25 @@ class Server(resource.Resource):
...
         migration_params = obj.migration_params()
+        from_image = params.get('image_ref')
+
+        # booted from local boot disk, boot-disk not copied
+        if migration_params['data_copy'] is False and from_image is not None:
+            migration_params['boot_disk_copy'] = False
 
         # boot-from-volume servers must copy boot disk
-        if params.get('image_ref') is None:
+        if from_image is None:
```
jistr, line 88: We probably don't need to change anything here, as in from_sdk migration params will always be defaulted, so 'data_copy' will never be False.

cloudguruab:
I've made some minor changes to code to support multiple storage options. So far here are some of my steps:

- [ ] Update the `migration_param_defaults` dictionary to include the `uuid` key and other keys.
    ◦ `data_copy`: A boolean value that indicates whether data should be copied from the source cloud to the destination cloud.
    
    ◦ `boot_volume`: A dictionary that specifies the boot volume for the destination VM. The dictionary should have the following key; `uuid`: The UUID of the boot volume.
    
    ◦ `additional_volumes`: A list of dictionaries that specify additional volumes for the destination VM. Each dictionary should have the following key to start; `uuid`: The UUID of the additional volume.
- [ ] Update the `create()` method to use the `data_copy`, `boot_volume`, and `additional_volumes` parameters to create the destination VM.
- [ ] Update the `update_migration_params()` method to validate the `data_copy`, `boot_volume`, and `additional_volumes` parameters.
- [ ] Update the `update_sdk_params_block_device_mapping()` method to use the `data_copy`, `boot_volume`, and `additional_volumes parameters` to update the block device mapping for the destination VM.
- [ ] In the `import_workloads` role, skip the tasks that copy data from the source cloud to the destination cloud.
- [ ] In the `import_workloads` role, add a task to create the destination VM using the `data_copy`, `boot_volume`, and `additional_volumes parameters`.
- [ ] Test the updated code to make sure that it can migrate VMs with a variety of storage configurations.

cloudguruab:
Current error, is just a simple test against expected output from the unittest, I don't think altering the static responses in the test themselves is the way to go, is there a way to update the fixtures to help get the responses in sync?

os_migrate/roles/conversion_host/tasks/conv_hosts_inventory.yml
```diff
@@ -27,6 +27,7 @@
       os_migrate_conversion_client_cert: "{{ os_migrate_dst_client_cert|default(omit) }}"
       os_migrate_conversion_client_key: "{{ os_migrate_dst_client_key|default(omit) }}"
 
+# TODO-storage do we need to handle this at all, given if data_copy is false then we dont deploy a src CH ?
```
cloudguruab, line 30: @jistr want to confirm this as well, do we want to search for conversion host at all?

os_migrate/roles/export_workloads/tasks/main.yml
```diff
@@ -32,6 +32,7 @@
     name: "{{ item['id'] }}"
     migration_params:
       boot_disk_copy: "{{ os_migrate_workloads_boot_disk_copy }}"
+      data_copy: "{{ os_migrate_workloads_data_copy }}"
```
jistr, line 35: Ah yes that's a good solution to allow us to override the default for all exports, i actually forgot we had this for boot_disk_copy.

jistr (changes requested):

os_migrate/tests/unit/test_server.py
```diff
@@ -259,6 +259,69 @@ def test_block_device_mapping_boot_disk_nocopy(self):
...
+                'source_type': 'volume',
+            },
+            {
+                'boot_index': -1,
+                'uuid': 'uuid-extra2-volume',
+                'delete_on_termination': False,
+                'destination_type': 'volume',
+                'source_type': 'volume',
+            }
+        ])
```
jistr, line 323: Very nice unit tests!

os_migrate/roles/import_workloads/tasks/workload.yml
```diff
@@ -1,133 +1,135 @@
...
-      msg:
-        - "{{ prelim.server_name }} log file: {{ prelim.log_file }}"
-        - "{{ prelim.server_name }} progress file: {{ prelim.state_file }}"
-    when: prelim.changed
+    - name: show progress data
+      ansible.builtin.debug:
+        msg:
+          - "{{ prelim.server_name }} log file: {{ prelim.log_file }}"
+          - "{{ prelim.server_name }} progress file: {{ prelim.state_file }}"
+      when: prelim.changed and data.data_copy|default(true)
```
jistr, line 23: `prelim` may be undefined if it wasn't executed. So we should probably do something like:

```
when:
  - prelim is defined and prelim.changed
  - data.data_copy|default(true)
```

This "is defined" guard on usage of prelim will have to be done in the `prelim` uses in the rest of the code below too.

os_migrate/plugins/module_utils/server.py
```diff
@@ -128,9 +141,60 @@ def update_migration_params(self, params_dict):
...
+                    additional_volume_mapping = {
+                        "boot_index": -1,
+                        'uuid': volume['uuid'],
+                        'delete_on_termination': False,
+                        'destination_type': 'volume',
+                        'source_type': 'volume',
+                    }
+                    block_device_mapping.append(additional_volume_mapping)
+            del sdk_params['image_id']
+        elif image_id is not None:
```
jistr, line 183: The elif/else block belongs to `if migration_params['boot_volume']['uuid']`. All of that deals with how the machine boots. The `additional_volumes` block deals with optional volumes and is independent.

tests/e2e/tasks/tenant/run_workload.yml
```diff
@@ -24,6 +24,13 @@
         replace: "boot_disk_copy: true"
       when: set_boot_disk_copy|default(false)|bool
 
+    - name: set data copy
+      ansible.builtin.replace:
+        path: "{{ os_migrate_data_dir }}/workloads.yml"
+        regexp: "data_copy: false"
+        replace: "data_copy: true"
+      when: set_data_copy|default(true)|bool
```
jistr, line 32: This should be in the inverse -- we should be looking for "data_copy: true" and replacing it with "data_copy: false". So the variable name should probably be `set_no_data_copy` or `set_data_copy_false` or `unset_data_copy`.

tests/e2e/test_as_tenant.yml
```diff
@@ -118,6 +120,8 @@
     #
     # * boot_disk_copy: true
     #
+    # * data_copy: true
+    #
```
jistr, line 124: This part looks good but we should also add a 3rd workload migration that would set "data_copy: false"

cloudguruab:
@jistr I can't recall if we needed to do something with listing src conversion hosts if data copy is false, but I made some changes to handle a edge case that was presented in the CI. From my understanding here: https://docs.openstack.org/nova/latest/user/block-device-mapping.html - openstack doesnt support mapping images to local boot disks for objects that aren't currently reserved for the entry referring to the Glance image that the instance is being booted with. Which produced this error: 

```python
TASK [os_migrate.os_migrate.import_workloads : transfer volumes to destination] ***
ok: [localhost] => changed=false 
  block_device_mapping:
  - boot_index: -1
    delete_on_termination: false
    destination_type: volume
    device_name: vdb
    source_type: volume
    uuid: 0b3b880c-e52a-48e4-9c13-cc57f50b3d2c
  volume_map:
    /dev/vdb:
      bootable: false
      dest_dev: /dev/vdb
      dest_id: 0b3b880c-e52a-48e4-9c13-cc57f50b3d2c
      image_id: null
      name: osm_volume
      port: 49152
      progress: 100.0
      size: 1
      snap_id: null
      source_dev: /dev/vdb
      source_id: 419fe613-d27f-4d7c-a562-588fe007d9ab
      url: nbd://localhost:49152/33ab5dd8-64b4-4df7-a58d-c1ab94ca8f37
TASK [os_migrate.os_migrate.import_workloads : create destination instance] ****
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: openstack.exceptions.BadRequestException: BadRequestException: 400: Client Error for url: http://10.0.110.65:8774/v2.1/servers, Block Device Mapping is Invalid: Mapping image to local is not supported.
fatal: [localhost]: FAILED! => changed=false 
  module_stderr: |-
...
    openstack.exceptions.BadRequestException: BadRequestException: 400: Client Error for url: http://10.0.110.65:8774/v2.1/servers, Block Device Mapping is Invalid: Mapping image to local is not supported.
  module_stdout: ''
  msg: |-
    MODULE FAILURE
    See stdout/stderr for the exact error
  rc: 1
```

I'm not sure if we should default to a destination type of `volume` for images when data copy is false to support **Booted from image (local boot disk), optionally additional volumes attached**. So I'm thinking: 

- Adding `destination_storage_type` users should specify when data_copy is false

cloudguruab:
#618 work moved here to make sure branch name follows contribution guidelines `f/<branch-name>`

Issue #611, opened 2023-05-16
locks persist after a failed workload migration


cloudguruab:

jistr:
I tried some of the locking mechanism locally (the flock command with all its parameters [1] and how it behaves if some of the files are left over) and i think for "good run" the locks that are left over don't prevent subsequent runs. However, after stuck runs, the cleanup logic does have an issue.

### How the locks work

The 2 different lock files are probably as the original implementation was intended:

* The locks in the format of `/var/lock/v2v-migration-lock.lock` (and others in the same format **with** `.lock` at the end) are the actual locks that need to be cleaned up if they are left over. The locking mechanism here is "does this file exist or not?". This is our "main lock" and it is meant to persist between process executions, and as such it has to be cleaned up explicitly by `check_and_cleanup_lockfiles` [4].

* The locks in the format of `/var/lock/v2v-migration-lock` (and others in the same format **without** `.lock` at the end) are only used by the `flock` command [1] and are only considered "locked" when the flock command is running. This is likely only to prevent a race condition when creating the other lock (`v2v-migration-lock.lock`). The locking mechanism for `v2v-migration-lock` is "is there a running process that has the file open?". That means crashed executions of a migration will never cause this lock to be "stuck in a locked state", and we don't need to clean up this file.


### Issues that i've noticed

* Sometimes the lock-checking loop [2] went through very fast. The loop is meant to pause 1 second between each iteration, but there are code paths that run "as fast as possible", and that has the effect that the whole loop completes in a much shorter time than intended, and we don't wait for a lock freeing up for as long as we should. We should fix this by putting `time.sleep(1)` before line 86 [3]

* The `check_process` and `check_and_cleanup_lockfiles` functions [4] do their work on the migrator host, but they should be doing their work on the conversion host. That's why they don't work as intended, they're trying to clean up those locks on a machine where the locks never get created. We should move these functions into the OpenStackHostBase class [5] and use `cmd_out` or `cmd_val` [6] for anything that we want to run on the conversion host itself (reading PID from file, checking running PID, removing the lock file etc.). This means we can't use pure Python code to perform those actions, we'll need to reformulate them as shell commands so that they can be executed over ssh.

[1] https://github.com/os-migrate/os-migrate/blob/385fec8842d84e8db968cc30b42915ccf88781d3/os_migrate/plugins/module_utils/workload_common.py#L75
[...]

PR #613, opened 2023-05-18
dev: lockfile bug for cleanup

cloudguruab:
Seems to be an issue with authentication within clouds, https://access.redhat.com/support/cases/#/case/03479904. Investigating issue, 

- [x] Request a machine with > `150GB`
- [x] Spin up a E2E environment to test
- [x] Test changes made from `workload_common`

### DIFF
--- a/os_migrate/plugins/module_utils/workload_common.py
+++ b/os_migrate/plugins/module_utils/workload_common.py
@@ -46,6 +46,7 @@ def check_and_cleanup_lockfiles():
     If a lockfile is not being used by a process,
     it is either deleted or not being used.
     """
+    # FIXME: bug in checking for locks that sometimes are missing .lock extension
     # Remove specific lockfiles
     for lockfile in [ATTACH_LOCK_FILE_SOURCE, ATTACH_LOCK_FILE_DESTINATION, PORT_LOCK_FILE]:
         try:
--- a/os_migrate/plugins/module_utils/workload_common.py
+++ b/os_migrate/plugins/module_utils/workload_common.py
@@ -32,12 +32,12 @@ def use_lock(lock_file):
     """ Boilerplate for functions that need to take a lock. """
     def _decorate_lock(function):
         def wait_for_lock(self):
-            time.sleep(1)
-            if check_and_cleanup_lockfiles():
+            if self.check_and_cleanup_lockfiles():
                 self.log.info("Cleaned up unused lockfiles before start")
             for second in range(DEFAULT_TIMEOUT):
                 try:
                     self.log.info('Waiting for lock %s...', lock_file)
+                    time.sleep(1)
                     lock = lock_file + '.lock'
                     pid = str(os.getpid())
                     cmd = ['sudo', 'flock', '--timeout', '1',
@@ -53,7 +53,6 @@ def use_lock(lock_file):
                         break
                 except subprocess.CalledProcessError as err:
                     self.log.info('Error waiting for lock: %s', str(err))
-                    time.sleep(1)
             else:
                 raise RuntimeError('Unable to acquire lock ' + lock_file)
             try:
@@ -118,7 +117,7 @@ class OpenStackHostBase():
 
     def check_process(self, pid):
         """Check whether pid exists in the current process table."""
-        if pid is None:
+        if not pid:
             return False
         try:
             self.shell.cmd_val(['kill', '-0', str(pid)])
@@ -134,18 +133,19 @@ class OpenStackHostBase():
         it is either deleted or not being used.
         """
         # Remove specific lockfiles
-        for lockfile in [ATTACH_LOCK_FILE_SOURCE, ATTACH_LOCK_FILE_DESTINATION, PORT_LOCK_FILE]:
-            try:
-                # FIXME cant use python modules here on migrator - has to be reworked
-                with open(f"{lockfile}.lock", "r", encoding='utf-8') as f:
-                    pid = f.read().strip() 
+        try:
+            # Remove specific lockfiles
+            for lockfile in [ATTACH_LOCK_FILE_SOURCE, ATTACH_LOCK_FILE_DESTINATION, PORT_LOCK_FILE]:
+                # Use self.shell.cmd_out to run the following command on the conversion host
+                pid = self.shell.cmd_out(['cat', f'{lockfile}.lock'])
                 if self.check_process(pid):
-                    pass
-                else:
-                    self.shell.cmd_val(['sudo', 'rm', '-f', lockfile + '.lock'])
-                    return True
-            except FileNotFoundError:
-                return False
+                    continue
+
+                # The lockfile is not being used by a process, so we can remove it
+                self.shell.cmd_val(['sudo', 'rm', '-f', lockfile + '.lock'])
+                return True
+        except FileNotFoundError:
+            return False
 
     def _converter(self):
         """ Refresh server object to pick up any changes. """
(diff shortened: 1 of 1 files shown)


cloudguruab:
converting python modules to scripts for conversion host, this will move search/cleanup off of migrators and into the proper channels.

os_migrate/plugins/module_utils/workload_common.py
```diff
@@ -148,6 +116,40 @@ def __init__(self, openstack_connection, conversion_host_id, ssh_key_path,
...
+            try:
+                # Remove specific lockfiles
+                for lockfile in [ATTACH_LOCK_FILE_SOURCE, ATTACH_LOCK_FILE_DESTINATION, PORT_LOCK_FILE]:
+                # Use self.shell.cmd_out to run the following command on the conversion host
+                pid = self.shell.cmd_out(f'cat {lockfile}.lock | grep -Po "(?<=^PID )\d+"')
+                if self.check_process(pid):
+                    continue                
+                else:
+                    # The lockfile is not being used by a process, so we can remove it
+                    # self.shell.cmd_out(f'rm {lockfile}.lock')
```
cloudguruab, line 147: @jistr I assume we're running as root user to begin with, if so I'll uncomment this command if not the command below will execute - users will need to run --ask-pass to get this to work so I can see a problem here

os_migrate/plugins/module_utils/workload_common.py
```diff
@@ -28,44 +28,12 @@
...
-        except FileNotFoundError:
-            return False
-
-
 def use_lock(lock_file):
     """ Boilerplate for functions that need to take a lock. """
     def _decorate_lock(function):
         def wait_for_lock(self):
-            if check_and_cleanup_lockfiles():
+            time.sleep(1)
```
jistr, line 35: Putting time.sleep(1) here will make the function always block for one second at least, but it will not fix the problem. The problem is with the `for second in range(DEFAULT_TIMEOUT):` below, one of the possible fixes should be what i suggested in #611 .

os_migrate/plugins/module_utils/workload_common.py
```diff
@@ -148,6 +116,39 @@ def __init__(self, openstack_connection, conversion_host_id, ssh_key_path,
...
+        Removes lockfiles found in specific directories.
+        If a lockfile is not being used by a process,
+        it is either deleted or not being used.
+        """
+        # Remove specific lockfiles
+        try:
+            # Remove specific lockfiles
+            for lockfile in [ATTACH_LOCK_FILE_SOURCE, ATTACH_LOCK_FILE_DESTINATION, PORT_LOCK_FILE]:
+                # Use self.shell.cmd_out to run the following command on the conversion host
+                pid = self.shell.cmd_out(f'cat {lockfile}.lock | grep -Po "(?<=^PID )\\d+"')
```
jistr, line 141: Is the grep needed/correct? Looking at this [1] i think the file just contains a number?

[1] https://github.com/os-migrate/os-migrate/blob/385fec8842d84e8db968cc30b42915ccf88781d3/os_migrate/plugins/module_utils/workload_common.py#L78

os_migrate/plugins/module_utils/workload_common.py
```diff
@@ -148,6 +116,39 @@ def __init__(self, openstack_connection, conversion_host_id, ssh_key_path,
         # Ports chosen for NBD export
         self.claimed_ports = []
 
+    def check_process(self, pid):
+        """Check whether pid exists in the current process table."""
+        if pid is None:
```
jistr, line 121: If we don't do input validation before calling the method (below on line 142), maybe we should handle getting an empty string here too, just in case the .lock is from an invocation of old OSM which didn't yet write the PID there? Perhaps

```
if not pid:
    return False
```

jistr (changes requested):

cloudguruab:
Was able to reproduce a similar [error](https://gitlab.cee.redhat.com/os-migrate/bot/-/jobs/12626827/raw) to the consultants: 

```
line 381, in cmd_out
      File "/usr/lib64/python3.8/subprocess.py", line 415, in check_output
        return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
      File "/usr/lib64/python3.8/subprocess.py", line 516, in run
        raise CalledProcessError(retcode, process.args,
    subprocess.CalledProcessError: Command '['ssh', '-o', 'BatchMode=yes', '-o', 'StrictHostKeyChecking=no', '-o', 'ConnectTimeout=10', '-i', '/root/os_migrate/tests/e2e/tmp/data/conversion/ssh.key', 'centos@172.19.13.148', 'c', 'a', 't', ' ', '/', 'v', 'a', 'r', '/', 'l', 'o', 'c', 'k', '/', 'v', '2', 'v', '-', 's', 'o', 'u', 'r', 'c', 'e', '-', 'v', 'o', 'l', 'u', 'm', 'e', '-', 'l', 'o', 'c', 'k', '.', 'l', 'o', 'c', 'k']' returned non-zero exit status 127.
  module_stdout: ''
  msg: |-
    MODULE FAILURE
    See stdout/stderr for the exact error
  rc: 1
```

os_migrate/plugins/module_utils/workload_common.py
```diff
@@ -28,54 +28,22 @@
...
     """ Boilerplate for functions that need to take a lock. """
     def _decorate_lock(function):
         def wait_for_lock(self):
-            if check_and_cleanup_lockfiles():
+            if self.check_and_cleanup_lockfiles():
                 self.log.info("Cleaned up unused lockfiles before start")
             for second in range(DEFAULT_TIMEOUT):
                 try:
                     self.log.info('Waiting for lock %s...', lock_file)
+                    time.sleep(1)
```
jistr, line 40: This will wait always, but we only need to wait if the first attempt fails and we need to make another attempt. I think the right spot for this waiting is right before the line `except subprocess.CalledProcessError as err`. (It should not be a part of the `elif result == 0` block though.)

os_migrate/plugins/module_utils/workload_common.py
```diff
@@ -148,6 +115,43 @@ def __init__(self, openstack_connection, conversion_host_id, ssh_key_path,
...
+        """
+        # Remove specific lockfiles
+        try:
+            # Remove specific lockfiles
+            for lockfile in [ATTACH_LOCK_FILE_SOURCE, ATTACH_LOCK_FILE_DESTINATION, PORT_LOCK_FILE]:
+                try:
+                    # Use self.shell.cmd_out to run the following command on the conversion host
+                    pid = self.shell.cmd_out(['sudo', 'cat', f'{lockfile}.lock'])
+                except Exception as err:
+                    pid = self.shell.cmd_out(['sudo', 'cat', f'{lockfile}'])
```
jistr, line 143: This file will never contain the PID, we only write it into 'lockfile.lock'. So i think this line can be dropped.

jistr (changes requested):
Two small things that i think should be tweaked, otherwise LGTM.

os_migrate/plugins/module_utils/workload_common.py
cloudguruab: I thought we wanted to place the time.sleep before line 86?

os_migrate/plugins/module_utils/workload_common.py
```diff
@@ -148,6 +115,43 @@ def __init__(self, openstack_connection, conversion_host_id, ssh_key_path,
...
+        """
+        Removes lockfiles found in specific directories.
+        If a lockfile is not being used by a process,
+        it is either deleted or not being used.
+        """
+        # Remove specific lockfiles
+        try:
+            # Remove specific lockfiles
+            for lockfile in [ATTACH_LOCK_FILE_SOURCE, ATTACH_LOCK_FILE_DESTINATION, PORT_LOCK_FILE]:
+                try:
```
jistr, line 139: From CI run it looks like we'll need `pid = None` above this.

jistr (approved):
LGTM, thanks!

PR #616, opened 2023-06-08
new: stable release for 0.15.1 galaxy

cloudguruab:


jistr:
I think this should be a PR against the stable/0.15 branch, and the stable/0.15.1 branch shouldn't exist :). We can chat about the details on 1-1.

PR #618, opened 2023-06-16
new storage migration

cloudguruab:
Moving work from #612 to this PR so I'm following contribution guidelines. Updates from last CI run, progress on the storage migration without conversion host changes:

1) Made changes from suggested comments on #612 
2) Running into a bug with migration scenario data_copy false, boot_disk_copy false, no additional volumes, no boot volume and a boot volume mapping for an image to local. CI Run Link: https://gitlab.cee.redhat.com/os-migrate/bot/pipelines/4341788

- [x] update unittests for any code changes
- [x] handle Booted from image (local boot disk), optionally additional volumes attached use-case
- [x] interactive rebase to cleanup commit history
- [x] update documentation - `variables_guide`, `migration params guide`, `how it works`, `role defaults`

### DIFF
--- a/os_migrate/plugins/module_utils/server.py
+++ b/os_migrate/plugins/module_utils/server.py
@@ -303,7 +303,7 @@ class Server(resource.Resource):
             conn, params['flavor_ref'])
 
         refs['image_ref'] = params['image_ref']
-        if self.migration_params()['boot_disk_copy']:
+        if self.migration_params()['data_copy'] and self.migration_params()['boot_disk_copy']:
             refs['image_id'] = None
         else:
             refs['image_id'] = reference.image_id(conn, params['image_ref'])
--- a/os_migrate/plugins/module_utils/server.py
+++ b/os_migrate/plugins/module_utils/server.py
@@ -60,14 +60,23 @@ class Server(resource.Resource):
 
     migration_param_defaults = {
         'boot_disk_copy': False,
+        'floating_ip_mode': 'auto',
         'boot_volume_params': {
             'availability_zone': None,
             'name': None,
             'description': None,
             'volume_type': None,
         },
-        'floating_ip_mode': 'auto',
-        'port_creation_mode': 'nova'
+        'port_creation_mode': 'nova',
+        'data_copy': True,
+        'boot_volume': {
+            'uuid': None,
+        },
+        'additional_volumes': [
+            {
+                'uuid': None,
+            }
+        ],
     }
 
     @classmethod
@@ -95,7 +104,11 @@ class Server(resource.Resource):
         # support for advanced port creation via Neutron.
         self.update_sdk_params_networks_simple(conn, sdk_params, port_creation_mode)
 
-        self.update_sdk_params_block_device_mapping(sdk_params, block_device_mapping)
+        if not self.migration_params()['data_copy']:
+            self.update_sdk_params_block_device_mapping_nocopy(sdk_params)
+        else:
+            self.update_sdk_params_block_device_mapping_copy(sdk_params, block_device_mapping)
+
         sdk_srv = conn.compute.create_server(**sdk_params)
         # Wait for the server before we attach Floating IPs, otherwise
         # we may hit "Instance network is not ready yet" error.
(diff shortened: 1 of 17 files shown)


os_migrate/tests/unit/test_server.py
```diff
@@ -259,6 +259,86 @@ def test_block_device_mapping_boot_disk_nocopy(self):
         ])
         self.assertTrue('image_id' not in sdk_params)
 
+    # TODO: add testcase for no data copy, no boot copy, load from optional image
```
cloudguruab, line 262: I've added a test case below to match the condition in the CI when boot-volume, additional volumes is empty, boot-disk-copy and data-copy are false like shown here: 

```
 _migration_params:
        additional_volumes:
        - uuid: null
        boot_disk_copy: false
        boot_volume:
          uuid: null
        boot_volume_params:
          availability_zone: null
          description: null
          name: null
          volume_type: null
        data_copy: false
        floating_ip_mode: auto
        port_creation_mode: neutron
```

os_migrate/plugins/module_utils/server.py
```diff
@@ -128,9 +141,60 @@ def update_migration_params(self, params_dict):
...
+                if volume['uuid']:
+                    additional_volume_mapping = {
+                        "boot_index": -1,
+                        'uuid': volume['uuid'],
+                        'delete_on_termination': False,
+                        'destination_type': 'volume',
+                        'source_type': 'volume',
+                    }
+                    block_device_mapping.append(additional_volume_mapping)
+            del sdk_params['image_id']
```
jistr, line 195: This 'del' should go between lines 169 and 170 so that we make sure we don't try to use an image if we have a boot volume.

jistr: And it should either be guarded by an if or it should be replaced with 'pop' like:  https://stackoverflow.com/questions/15411107/delete-a-dictionary-item-if-the-key-exists

os_migrate/tests/unit/test_server.py
```diff
@@ -259,23 +260,14 @@ def test_block_device_mapping_boot_disk_nocopy(self):
...
-                'boot_index': 0,
-                'uuid': 'uuid-test-image',
-                'delete_on_termination': False,
-                'destination_type': 'local',
-                'source_type': 'image',
-            }
-        ])
-        self.assertEqual(sdk_params['image_id'], 'uuid-test-image')
+        with self.assertRaises(exc.InconsistentState):
+            srv.update_sdk_params_block_device_mapping_nocopy(sdk_params)
```
jistr, line 270: IMO it wold make sense to have both of these tests -- both for when image is specified and when nothing is specified and we need to raise an error. Now that the removal of 'image_id' from sdk_params is fixed, i think the original unit test shouldn't fail anymore.

image

os_migrate/roles/import_workloads/tasks/workload.yml
```diff
@@ -1,133 +1,149 @@
...
+        ca_cert: "{{ os_migrate_dst_ca_cert|default(omit) }}"
+        client_cert: "{{ os_migrate_dst_client_cert|default(omit) }}"
+        client_key: "{{ os_migrate_dst_client_key|default(omit) }}"
+        dst_filters: "{{ os_migrate_dst_filters }}"
+        src_conversion_host:
+          "{{ os_src_conversion_host_info.openstack_conversion_host }}"
+        log_dir: "{{ os_migrate_data_dir }}/workload_logs"
+        data: "{{ item }}"
+      register: prelim
+      when: data.data_copy|default(true)
```
jistr, line 16: I'd try `item["_migration_params"]["data_copy"]|default(true)`

os_migrate/roles/import_workloads/tasks/workload.yml
```diff
@@ -1,133 +1,149 @@
...
+        validate_certs: "{{ os_migrate_dst_validate_certs|default(omit) }}"
+        ca_cert: "{{ os_migrate_dst_ca_cert|default(omit) }}"
+        client_cert: "{{ os_migrate_dst_client_cert|default(omit) }}"
+        client_key: "{{ os_migrate_dst_client_key|default(omit) }}"
+        data: "{{ item }}"
+        block_device_mapping: "{{ transfer.block_device_mapping }}"
+      register: os_migrate_destination_instance
+      when:
+        - prelim is defined and prelim.changed
+        - item["_migration_params"]["data_copy"]|default(true)
```
jistr, line 126: The e2e test path with `data_copy: false` skipped the creation of the instance because of this:

```
TASK [os_migrate.os_migrate.import_workloads : create destination instance] ****
skipping: [localhost] => changed=false 
  skip_reason: Conditional result was False
```

I think we should keep the prelim condition but only for the case when `data_copy: true` (because when running with `data_copy: false`, `prelim` will be undefined).

So i think we should have:
```
when:
  - prelim is defined and prelim.changed or not item["_migration_params"]["data_copy"]|default(true)
```

That condition should evaluate to true (task will not be skipped) when:

* item["_migration_params"]["data_copy"] is false, and then we don't care about prelim, or
* item["_migration_params"]["data_copy"] is true, and prelim is defined and it's changed.

jistr (changes requested):

os_migrate/plugins/module_utils/server.py
```diff
@@ -128,9 +141,60 @@ def update_migration_params(self, params_dict):
...
+            boot_vol_mapping = {
+                "boot_index": 0,
+                "uuid": migration_params['boot_volume']['uuid'],
+                'delete_on_termination': False,
+                'destination_type': 'volume',
+                'source_type': 'volume',
+            }
+            block_device_mapping.append(boot_vol_mapping)
+            sdk_params.pop("image_id", None)
+        elif image_id is not None:
```
jistr, line 171: I think this is still None because an earlier piece of logic for copying the boot disks is kicking in here and setting the image_id to None, here:

https://github.com/os-migrate/os-migrate/blob/ba207d3eb909d075b453d6c2a86fe58498c3c7e4/os_migrate/plugins/module_utils/server.py#L306

We should be setting the image_id to none only if we are doing data_copy in general *and* specifically if we're doing also boot_disk_copy. So i think the condition there should be:

```
if self.migration_params()['data_copy'] and self.migration_params()['boot_disk_copy']:
```

tests/e2e/tasks/tenant/run_workload.yml
```diff
@@ -24,6 +24,13 @@
         replace: "boot_disk_copy: true"
       when: set_boot_disk_copy|default(false)|bool
 
+    - name: set data copy
+      ansible.builtin.replace:
+        path: "{{ os_migrate_data_dir }}/workloads.yml"
+        regexp: "data_copy: true"
+        replace: "data_copy: false"
+      when: set_no_data_copy|default(true)|bool
```
jistr, line 32: I think this behaves inverse to what was intended and what the comments in workload migration tests say. We should default to not doing anything here (to keep existing e2e workload tests working the same way).

So i think we should have:

```
when: set_no_data_copy|default(false)|bool
```

tests/e2e/test_as_tenant.yml
```diff
@@ -143,6 +196,7 @@
       tags: always
       vars:
         set_boot_disk_copy: true
+        set_no_data_copy: false
```
jistr, line 199: And here it should be `set_no_data_copy: true`. Then i think the tests will behave as the comments here say.

jistr (commented):
We should fix the e2e test matrix behavior, otherwise LGTM :+1:

ccamacho (approved):

Issue #619, opened 2023-07-17
Cannot connect to newly deployed UCH hosts


njnatale-lm:
I'm using the CentOS 9 cloud image linked in the documentation, but seem to hit a few issues (this is with the latest version of `os-migrate`):

First, when I deploy the host, I get this at the end:

```
TASK [Gathering Facts] **********************************************************************************************************************************************
fatal: [os_migrate_conv_src]: UNREACHABLE! => {"changed": false, "msg": "Failed to connect to the host via ssh: Warning: Permanently added 'XX.XX.XX.XX' (ECDSA) to the list of known hosts.\r\nPermission denied (publickey,gssapi-keyex,gssapi-with-mic).", "unreachable": true}
```

The conversion host's IP was masked by me for the purpose of reporting this; I am able to ping this IP, but can't get in.

I wanted to debug further, so I set `os_migrate_conversion_host_ssh_user_enable_password_access` to `true` - but I still cannot SSH in, and do not get a password prompt.

I'm not super familiar with CentOS 9, but I'm wondering if password-based SSH is disabled by default; though that doesn't really help explain why Ansible can't get in with the keys it generated a few minutes earlier in the deployment process.

Any help on both of these related issues would be appreciated!

njnatale-lm:
Ah, I think my issue was that I still was setting the username of the UCH account to `cloud-user` not realizing that the latest recommended image uses `centos` (the previous image from a year or so ago used `cloud-user`).

jistr:
Thank you for the report, if you don't mind i'll reopen the issue, because i think we should update the defaults in the variables files. I think the defaults in this project should assume we're using latest CentOS Stream as the conversion host OS.

------

For our reference:

https://github.com/os-migrate/os-migrate/blob/93d1a7d7dc7ea0e44fb4df4bcfb559bc96784d3e/os_migrate/roles/conversion_host/defaults/main.yml#L34

https://github.com/os-migrate/os-migrate/blob/93d1a7d7dc7ea0e44fb4df4bcfb559bc96784d3e/os_migrate/roles/conversion_host_content/defaults/main.yml#L16

njnatale-lm:
I think this should now be good to close.

@cloudguruab the only question I have- is there a bug in what generates the [changelog](https://os-migrate.github.io/os-migrate/changelog.html)? I'm a little curious at why it represents my name in two different ways for the two different PRs. Not a big deal, just curious 😄

njnatale-lm:
[...]

PR #625, opened 2023-07-24
fix: set default username to centos

njnatale-lm:
change default username for CentOS 9 cloud image (#619)


ccamacho (approved):

njnatale-lm:
Looks much better now, thank you @cloudguruab! 👍

PR #636, opened 2023-08-10
upgrade venv-requirements in toolbox

cloudguruab:
This PR is a draft of work in #631 - goal is to segment which parts we're upgrading to handle errors. 

- [x] update other requirement files
- [x] patch to missing auth issue for setting tags in `subnets`, `routers`, `networks` functional tests
- [x] documentation updates
- [x] walk through `security_group_rules` to clean from testcases
- [x] solidify ci-matrix
- [x] update default user for cloud back to centos

### DIFF
--- a/os_migrate/plugins/module_utils/workload_common.py
+++ b/os_migrate/plugins/module_utils/workload_common.py
@@ -210,7 +210,7 @@ class OpenStackHostBase():
             new_disks = disks_after - disks_before
             volume = conn.get_volume_by_id(volume_id)
             attachment = self._get_attachment(volume, host_func())
-            dev_path = attachment.device
+            dev_path = attachment['device']
             if len(new_disks) == 1:
                 if dev_path in new_disks:
                     self.log.debug('Successfully attached new disk %s, and %s '
@@ -221,7 +221,7 @@ class OpenStackHostBase():
                     self.log.debug('Successfully attached new disk %s, but %s '
                                    'conversion host path does not match the  '
                                    'result from OpenStack. Using internal '
-                                   'device path %s.', attachment.device,
+                                   'device path %s.', attachment['device'],
                                    name, dev_path)
             else:
                 raise RuntimeError('Got unexpected disk list after attaching '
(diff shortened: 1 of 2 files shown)


os_migrate/roles/prelude_common/defaults/main.yml
```diff
@@ -14,4 +14,4 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
-os_migrate_common_minimum_openstacksdk_version: "0.36.0"
+os_migrate_common_minimum_openstacksdk_version: "0.99.0"
```
jistr, line 17: I'd go with 1.0 here, like what we have above in the pip install requirements.

os_migrate/galaxy.yml
```diff
@@ -19,7 +19,7 @@ tags:
 dependencies:
   community.crypto: '>=1.4.0'
   community.general: '<5.0.0'
-  openstack.cloud: '<2.0.0'
+  openstack.cloud: '==2.0.0'
```
jistr, line 22: I think we may want go unpin and go with `>=2.0.0` and only pin to a specific version if we find that there is something that breaks us? I think the SDK should now be pretty stable and i'd hope that would somewhat translate to the Ansible collection too...

toolbox/build/venv-requirements-ansible-2.9.txt
jistr, line 1: I'm a bit puzzled why this removal doesn't break the e2e CI. I think removing those files and simplifying the CI matrix is good, but i thought we were using these in the e2e so we may want to double check the config there later, just to make sure we're still testing what we intend to.

jistr:
In the 16 cloud with the new SDK, the default egress rules don't get created, we'll need to amend the os_migrate_conv security group rules creation [1] to include egress:

```
+--------------------------------------+-------------+-----------+-----------+------------+-----------+--------------------------------------+----------------------+
| ID                                   | IP Protocol | Ethertype | IP Range  | Port Range | Direction | Remote Security Group                | Remote Address Group |
+--------------------------------------+-------------+-----------+-----------+------------+-----------+--------------------------------------+----------------------+
| cdfd69ca-c0b0-4d96-bf0e-525702e1910f | None        | IPv4      | 0.0.0.0/0 |            | egress    | None                                 | None                 |
| eee4be98-0e39-48e6-bddc-7e8f0873a745 | None        | IPv6      | ::/0      |            | egress    | None                                 | None                 |

```

[1] https://github.com/os-migrate/os-migrate/blob/426121f307bf0078c1c72b6dd48980849db2c768/os_migrate/roles/conversion_host/tasks/main.yml#L23-L45

PR #650, opened 2023-11-27
fix: hot migration conversion host functionality for data copy

cloudguruab:


docs/src/user/variables-guide.rst
```diff
@@ -459,3 +459,129 @@ Workload import/export variables
...
+   :widths: 25 25 25 25
+   :header-rows: 1
+
+   * - Source workload
+     - Desired migration properties
+     - Migration parameter settings
+     - Conversion hosts required?
+   * - Booted from image (local boot disk), optionally additional volumes attached
+     - Boot disk not copied, VM booted from image (local boot disk), any additional volumes copied and attached
+     - _migration_params:
```
jistr, line 484: These code blocks don't get rendered correctly, i wonder if we could do a block enclosed with '```' inside the table cell? But in general ++ on having the table here.

https://github.com/os-migrate/os-migrate/blob/2ad62a288fc11698b50472388d35b2d8fce4a701/docs/src/user/variables-guide.rst#workload-migration-variables

os_migrate/roles/conversion_host/tasks/conv_hosts_inventory.yml
```diff
@@ -26,22 +26,30 @@
       os_migrate_conversion_ca_cert: "{{ os_migrate_dst_ca_cert|default(omit) }}"
       os_migrate_conversion_client_cert: "{{ os_migrate_dst_client_cert|default(omit) }}"
       os_migrate_conversion_client_key: "{{ os_migrate_dst_client_key|default(omit) }}"
+  when:
+    - item["_migration_params"]["data_copy"]|default(true)
```
jistr, line 30: The `item` here is a conversion host. It will not have `_migration_params` on it like the workloads do, so i think what's happening here is that `default(true)` gets used always. We will have to think of some other solution. Maybe we should not include this file at all if `os_migrate_workloads_data_copy` is `false`?

jistr (changes requested):

docs/src/user/variables-guide.rst
jistr: I meant for specific elements in the table. But the code looks good and CI is green, so let's merge and fix up the docs in a follow up.

jistr (approved):

PR #654, opened 2023-12-11
dev: patch to error output for missing resource file

cloudguruab:
Improved error output shown here: 

```
TASK [os_migrate.os_migrate.validate_resource_files : validate resource files] ***
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: ansible_collections.os_migrate.os_migrate.plugins.module_utils.exc.EmptyYAMLFileError: 
Detected empty resource file at /root/os_migrate/tests/e2e/tmp/data/networks.yml
```

### DIFF
--- a/os_migrate/plugins/module_utils/filesystem.py
+++ b/os_migrate/plugins/module_utils/filesystem.py
@@ -38,11 +38,10 @@ def load_resources_file(file_path):
     with open(file_path, 'r', encoding='utf8') as f:
         file_struct = yaml.safe_load(f)
 
-    if file_struct is not None:
-        file_os_migrate_version = file_struct.get('os_migrate_version', None)
+    if file_struct is None:
         raise exc.EmptyYAMLFileError(file_path)
-    else:
-        raise exc.EmptyYAMLFileError(file_path)
+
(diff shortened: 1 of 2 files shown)


os_migrate/plugins/module_utils/filesystem.py
```diff
@@ -38,7 +38,11 @@ def load_resources_file(file_path):
     with open(file_path, 'r', encoding='utf8') as f:
         file_struct = yaml.safe_load(f)
 
-    file_os_migrate_version = file_struct.get('os_migrate_version', None)
+    if file_struct is not None:
+        file_os_migrate_version = file_struct.get('os_migrate_version', None)
+    else:
+        raise exc.EmptyYAMLFileError(file_path)
```
jistr, line 44: This code block can be simplified to:

```
    if file_struct is None:
        raise exc.EmptyYAMLFileError(file_path)

    file_os_migrate_version = file_struct.get('os_migrate_version', None)
```

Basically, we check for the exception condition first, if it's present then we raise the exception. Then we continue with the happy path as usual. Not a blocker but it's a customary way to deal with error conditions, to keep the control flow simple, so it may be worth an update perhaps...

jistr (approved):

PR #666, opened 2024-04-25
dev: openstacksdk bump to 1.0.0 support with collections 2.x.x

cloudguruab:
we'll need to update the bot repo as well for pipelines to use the standard reqs file so we can purge the lower-bound.


fdiazbra (approved):
lgtm

PR #674, opened 2024-06-24
d/cleanup docs update

cloudguruab:
Update documentation between user guides and gitlab.

### DIFF
--- a/docs/src/user/variables-guide.rst
+++ b/docs/src/user/variables-guide.rst
@@ -472,7 +472,7 @@ is the conversion host required for this mode of migration, is this migration
 mode implemented).
 
 .. list-table:: Workload Migration options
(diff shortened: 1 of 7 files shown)


docs/src/user/variables-guide.rst
```diff
@@ -511,7 +511,7 @@ mode implemented).
         name: some_image
         project_name: '%auth%'
 
-        (image_ref doesn’t matter, can be null)```
+        (image_ref doesnt matter, can be null)```
```
davidjpeacock, line 514: typo - "doesn't" would be needed here, or "does not" would be more formal

davidjpeacock (changes requested):
Need a few things fixed up, comments inline

fdiazbra (approved):
LGTM

PR #692, opened 2024-09-13
Set cloud_vars_src fact

matbu:
Set cloud_vars_src fact even when os_migrate_src_block_storage_api_version is not defined otherwise the next task failed while trying to reused the cloud_vars_src variable.

### DIFF
--- a/os_migrate/roles/import_workloads/tasks/workload.yml
+++ b/os_migrate/roles/import_workloads/tasks/workload.yml
@@ -77,7 +77,6 @@
         cloud_vars_src: "{{ cloud_vars.clouds['src'] | combine({'block_storage_api_version' : '2'}) }}"
       when:
         - prelim is defined and prelim.changed
-        - os_migrate_src_block_storage_api_version is defined
         - item["_migration_params"]["data_copy"]|default(true)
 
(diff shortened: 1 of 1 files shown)


cloudguruab:
The ongoing issue in the functional test relates to [[1]](https://github.com/os-migrate/os-migrate/blob/main/tests/func/ansible.cfg)


```
ERROR! We were unable to read either as JSON nor YAML, these are the errors we got from each:
JSON: Expecting value: line 1 column 1 (char 0)

Syntax Error while loading YAML.
  mapping values are not allowed in this context

The error appears to be in '/root/os_migrate/tests/auth_tenant.yml': line 18, column 9, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

Building the toolbox container image
STEP 1/9: FROM quay.io/fedora/fedora:36-x86_64
        ^ here
make: *** [Makefile:65: test-func-tenant] Error 4
```

cloudguruab:
Not sure about the e2e ci failure, I think this is the connectivity issue from earlier?

matbu:
recheck

matbu:
recheck

PR #693, opened 2024-09-16
dev: patch to lockfile cleanup

cloudguruab:
- Process All Lockfiles: Removed the outer try-except block that could cause an early exit. The method now iterates over all lockfiles regardless of errors encountered with any single one.

- Specific Exception Handling: Changed except Exception to except subprocess.CalledProcessError for more precise error handling.

- Check Return Codes: If cat fails with exit code 1 (file not found), it's logged at the debug level and the loop continues to the next lockfile.

- Logging Improvements: Added more informative logging to help trace the flow and understand which lockfiles were missing, which were held, and which were removed.

- Correct Return Value: The method now only returns True if it actually cleaned up any lockfiles, ensuring that the calling function behaves correctly.

### DIFF
--- a/os_migrate/plugins/module_utils/workload_common.py
+++ b/os_migrate/plugins/module_utils/workload_common.py
@@ -133,26 +133,31 @@ class OpenStackHostBase():
         If a lockfile is not being used by a process,
         it is either deleted or not being used.
         """
-        # Remove specific lockfiles
-        try:
-            # Remove specific lockfiles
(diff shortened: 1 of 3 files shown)


cloudguruab:
- sudo  rm -rf  .local/share/containers/

cloudguruab:
Seems like the issue is related to something similar [here](https://github.com/containers/podman/issues/19604), `df -h` returns: 
```
[cloud-user@osm-ci-heat-runner ~]$ df /var/tmp/ 
Filesystem     1K-blocks     Used Available Use% Mounted on
/dev/sda1       31445996 24482140   6963856  78% /
```

PR #696, opened 2024-10-07
dev: Upgrade of the toolbox container and dependencies bump

rebtoor:
- The base image for the toolbox container has been updated to Fedora 40.
- The base image for Vagrant has been switched from Fedora 36 to CentOS Stream 9.
- Python packages have been updated and pinned to their latest stable versions.
- Official vagrant build will be used and vagrant-livirt plugin will be installed from upstream.
- Devstack is now configured on `2023.1` branch.


fdiazbra (approved):
Looks good to me!

PR #710, opened 2025-02-21
OSPRH-13551 add docs vmware section to readme

davidjpeacock:
This PR presents a new directory for our os-migrate GitHub.io single source of truth documentation, and represents the first of many changes to come in the documentation overhaul.

This imports the VMware docs from the foreign repo, with formatting changes for our Sphinx setup.


matbu:
Thank you for this PR.

Just a general comment, I don't think we should move the Vmware collection documentation into this repo:
* the galaxy web site reads the collection/README.rst doc into the repo, so the source for the galaxy doc will be the one into the vmware repo.
* due to ^ we will have to do manual sync each time the vmware/README is updated.
* it's duplacating content (and make out of sync, for example the content of the PR is already outdated).

So I think the solutions would be:
* to have a job that synced doc from the two collections in a third party repo (for example, just an idea)
* to have two differents documents for the two collections.
Also most generaly there is some topics which fit in the middle of the collections: building an AEE, setting up AAP for OSM and maybe some of "how to" or recorded demo on how to use each of those collections.

davidjpeacock:
No longer moving forward with this approach.

PR #715, opened 2025-03-10
Add ansible-lint workflow

matbu:


matbu:
In order to move forward to the certifications of the os-migrate collections, I add this ansible-lint job.

PR #716, opened 2025-03-10
dev: add ansible-lint github workflow

matbu:
Add ansible-lint job for Ansible certification.


rebtoor:
Hi @matbu , thanks for the PR! 

Rather than manually install `pip` and `ansible-lint`, what do you think about using the official github action provided by ansible folks? https://github.com/marketplace/actions/run-ansible-lint

(Thanks for having updated the ubuntu images!)

rebtoor (approved):

PR #717, opened 2025-03-24
dev: detached cinder volumes migration feature

fdiazbra:
Additions:
* Playbooks for the import/export of detached volumes
* Roles for the import/export of detached volumes
* Modules for import/export of detached volumes
* e2e test to include a detached volume migration
* Documentation files for new roles and modules for the autodoc
* Documentation and diagrams about the new workflow

Refactor:
* All methods in the various modules related to volumes are moved to the volumes_common.py to avoid code duplication
* Conversion hosts verifications moved to the role conv_host in the conv_host_details task
* Prepare_exports method set as abstract and implemented in each module since it has a different behaviour if we are migrating detached volumes or attached to a workload

### DIFF
--- a/docs/src/images/plantuml/detached-volume-migration-data-flow.plantuml
+++ b/docs/src/images/plantuml/detached-volume-migration-data-flow.plantuml
@@ -1,0 +1,44 @@
+@startuml
+
+participant Migrator as p_mig
+participant "Destination\nCloud" as p_dst_cloud
+participant "Destination\nConv. Host" as p_dst_host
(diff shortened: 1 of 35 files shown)


matbu (approved):
Thank you for this (very) big work. It's a bit hard to review and catch everything, but I think we should move on and fix nits itteratively.

davidjpeacock:
Merging this because review would be largely an exercise in code-reading, I have not tested, but as Mathieu suggests, I think it's important to get this merged and iterate on nits as they come up.

Great work Fernando!

PR #720, opened 2025-05-20
new: adding actions/steps for running ansible tests/sanity checks

cloudguruab:
1. Separated the build setup into a reusable **actions-build.yml** workflow
2. Simplified both **ansible-sanity.yml** and **ansible-unit.yml** to focus on their specific test purposes
3. Created a cleaner dependency structure with needs: build


.github/workflows/ansible-sanity.yml
```diff
@@ -0,0 +1,33 @@
...
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Download collection
+        uses: actions/download-artifact@v4
+        with:
+          name: ansible-collection
+          path: ansible_collections
+
+      - name: Run sanity checks
```
cloudguruab, line 28: The following + whats in unit workflows should fail intentionally.

os_migrate/galaxy.yml
```diff
@@ -18,9 +18,9 @@ tags:
   - openstack
   - migrations
 dependencies:
-  community.crypto: ">=1.4.0"
-  community.general: "<5.0.0"
   openstack.cloud: ">=2.3.0,<2.4.0"
+  community.crypto: ">=1.4.0"  # [COMMUNITY]
+  community.general: "<5.0.0"  # [COMMUNITY]
```
rebtoor, line 23: since we're refactoring here... why are we pinned `community.general` collection under 5.0.0? Last version is `10.6.0` so I'd give it a try.

rebtoor (changes requested):
To avoid to reinvent the wheel and have different behavior on repositories of the same org, I'd follow the same pattern that we adopted in vmware-migration-kit:
1. all the tests/linting are performed in a virtualenv environment
2. the tests are launched by `make` targets
3. the github actions consume the make targets

In this way you allow developers and users to launch the tests in the same way they are executed remotely without copy/paste any commands.

os_migrate/galaxy.yml
cloudguruab: Good catch, will do!

PR #725, opened 2025-07-01
Refactor of the os-migrate repo

rebtoor:
So... this is a major refactoring of the repo. I've tried to split the commits as much as i can, but due to the nature of the work itself, it was difficult.

Here's a summary of the things done:

- the collection directory was taken out the directory `os-migrate`, so now it's in the root folder
- the toolbox container has been removed, replaced by an AEE container which has another purpose
- functional tests have been removed from github actions because they need to assested (the actual way of launching them using a standalone deploy of devstack on a github runner is sub-optimal. This job is tracked in https://issues.redhat.com/browse/OSPRH-16827)
- all the targets in the Makefile have been rewritten to be launched in a CentOS Stream 10 containers and the github actions jobs have been adapted accordingly
- building of the collection has been greatly simplified
- linter errors and sanity errors are tracked in different cards and will be addressed in different PRs
- added github actions for galaxy import test and ansible tests (both sanity and units)
- ansible-lint has been ported from `min` profile to `production` profile

**BE aware that jobs are broken now:**

1. we need to address linting/sanity errors
2. we need to change how to launch the downstream CI jobs

so when the pr will be merged, it will leave the tests in a broken state for a while. We'll try to fix that as soon as we can.


davidjpeacock:
I'm actively reviewing this, but I'd like at least one more thorough review.  Thanks for the work Roberto, this is big!

fdiazbra (approved):
Looks good to me! Great job!

rebtoor:
> LGTM! We should keep track of doc changes in an issue (dont think it really needs a jira card unless it already exist) - with the repo change fragments are likely broken in small areas for user docs.

Already fixed in [this](https://github.com/os-migrate/os-migrate/pull/725/commits/0c81a4efaa8dca796dbe6bc5c4f7da103be55c72) commit mate :D

PR #740, opened 2025-10-03
Use CinderV3 by Default

sempervictus:
CinderV2 API has been gone since Xena so the probability of users accessing such old clouds is dropping daily.

Update to modern default in the workload migration task.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -173,30 +173,3 @@
       defend, and hold each Contributor harmless for any liability
       incurred by, or claims asserted against, such Contributor by reason
       of your accepting any such warranty or additional liability.
-
-   END OF TERMS AND CONDITIONS
-
-   APPENDIX: How to apply the Apache License to your work.
-
-      To apply the Apache License to your work, attach the following
-      boilerplate notice, with the fields enclosed by brackets "[]"
-      replaced with your own identifying information. (Don't include
-      the brackets!)  The text should be enclosed in the appropriate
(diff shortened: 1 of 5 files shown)


sempervictus:
While we're at it, shouldn't
```py
cloud_vars_src: "{{ cloud_vars.clouds['src'] | combine({'block_storage_api_version': '3'}) }}"
```
actually be 
```py
cloud_vars_src: "{{ cloud_vars.clouds['src'] | combine({'os_migrate_src_block_storage_api_version': '3'}) }}"
```
?

sempervictus:
For reference: this addresses a problem with migrating out of an impressively hosed openstack messed up to a level where it doesn't properly advertise the v3 endpoint causing clients to fall-back to defaults.

pczarnik:
version 3 is used already in `import_detached_volumes`:
https://github.com/os-migrate/os-migrate/blob/b05299e70b233a1cb217615bfc1ead2f44efd0ce/roles/import_detached_volumes/tasks/import_volumes.yml#L6-L8

Is this expected that `import_workloads` still uses version 2, while `import_detached_volumes` uses version 3?

pczarnik:
hi @fdiazbra, do you possibly remember why you used version 3 in #717?

fdiazbra:
> hi @fdiazbra, do you possibly remember why you used version 3 in #717?

Hi! I'd say that v2 was deprecated long time ago, as @sempervictus commented, so we should merge this in my opinion

PR #765, opened 2026-03-27
Add openstack.cloud as vendor collection instead of installing it fro…

matbu:
…m galaxy


Makefile
```diff
@@ -94,21 +95,84 @@ help:
...
+
+.PHONY: vendor-import vendor-links vendor-clean
+
+# Import vendor openstack.cloud collection.
+vendor-import: vendor-clean
+	@echo "--- Initializing OpenStack upstream submodule at version $(OS_CLOUD_VERSION) ---"
+	@mkdir -p $(VENDOR_DIR)
+	@if [ ! -d "$(UPSTREAM_REPO)" ]; then \
+		echo "Adding submodule..."; \
+		git submodule add https://github.com/openstack/ansible-collections-openstack.git $(UPSTREAM_REPO); \
```
cloudguruab, line 135: So submodules are workarounds for certification/collection specific dependencies? This is interesting 🚀

plugins/module_utils/osm_resource.py
cloudguruab, line 1: We're still using this file tho? 🤔 A bit confused on the submodule approach here, only because I havent seen it before. It seems cleaner than a git clone step from a separate repo for the tests folders.

cloudguruab (approved):
LGTM! Just have some context oriented questions

Makefile
matbu: @cloudguruab not really "workaround" but a way to include "validate" external dependencies into a project. Here we are pulling the modules from a specific stable version and only the modules we need and have tested.

plugins/module_utils/osm_resource.py
matbu: Yes we are still using the osm resource.py, but openstack.cloud has also a resource.py, that's why I had to rename it.
Basically, the git submodule will clone the openstack.cloud collection in a _vendor/ directory, then we pick up only a list of modules (and module_utils as well) and we linked them into our plugins/modules/ and plugins/module_utils/
So the os-migrate playbooks and roles, can resolve the openstack.cloud.foo modules to os_migrate.os_migrate.foo.
I will do a quick demo at the end of the sprint to be more clear about that before merging for every one.

PR #770, opened 2026-05-06
Remove community.crypto dependency

matbu:
This pull request removes the community.crypto dependency.
In order to certified OS-Migrate we need to cut dependency to community collections and use either builtin or certificed/validated collections.

### DIFF
--- a/roles/conversion_host/tasks/generate_keypair.yml
+++ b/roles/conversion_host/tasks/generate_keypair.yml
@@ -6,6 +6,8 @@
 
 - name: Generate a keypair for the conversion host
   ansible.builtin.shell: >
-    ssh-keygen -t rsa -b 4096 -f "{{ os_migrate_conversion_keypair_private_path }}" -N ""
+    ssh-keygen -t {{ os_migrate_conversion_host_ssh_key_type }}
+    {% if os_migrate_conversion_host_ssh_key_type != 'ed25519' %}-b {{ os_migrate_conversion_host_ssh_key_bits }}{% endif %}
+    -f "{{ os_migrate_conversion_keypair_private_path }}" -N ""
   args:
     creates: "{{ os_migrate_conversion_keypair_private_path }}"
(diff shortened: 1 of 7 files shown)


roles/conversion_host/tasks/generate_keypair.yml
```diff
@@ -5,5 +5,9 @@
...
 
 - name: Generate a keypair for the conversion host
-  community.crypto.openssh_keypair:
-    path: "{{ os_migrate_conversion_keypair_private_path }}"
+  ansible.builtin.user:
+    name: "{{ lookup('env', 'USER') }}"
+    generate_ssh_key: yes
+    ssh_key_bits: 4096
+    ssh_key_file: "{{ os_migrate_conversion_keypair_private_path }}"
+    ssh_key_type: rsa
```
rebtoor: missing EOF newline here!

rebtoor (approved):
Just a nit but ++

roles/conversion_host/tasks/generate_keypair.yml
matbu: erf thank you (vs code  :/ )

rebtoor (approved):

rebtoor:
@matbu i was thinking that probably using a more recent cipher like `edd2519` or `ecdsa` would be better, also I'd make cipher and its size customizable (with a default), WDYT?

See: https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html/configuring_basic_system_settings/assembly_using-secure-communications-between-two-systems-with-openssh_configuring-basic-system-settings

tests/e2e/tasks/tenant/seed_pre_workload.yml
```diff
@@ -104,10 +104,8 @@
     state: directory
 
 - name: Generate a keypair for the migration
-  # This will not regenerate the key if
-  # it already exists
-  community.crypto.openssh_keypair:
-    path: "{{ '~' | expanduser }}/ssh-ci/id_rsa"
+  ansible.builtin.shell: >
+    ssh-keygen -t ed25519 -f "{{ '~' | expanduser }}/ssh-ci/id_rsa }}" -N ""
```
rebtoor: Should it be `id_ed25519`? `ssh` command should find it as well as default.

fdiazbra (approved):
LGTM

PR #772, opened 2026-05-14
update checksum to usedforsecurity=false

cloudguruab:
MD5 verifies downloaded image data against Glance's Content-MD5 header a check, not authentication. The server dictates MD5, so SHA-256 isn't an option. `usedforsecurity=False` keeps this working in FIPS-mode Python.

### DIFF
--- a/plugins/module_utils/image.py
+++ b/plugins/module_utils/image.py
@@ -7,6 +7,7 @@ import os
 
 try:
     import openstack
+
     HAS_OPENSTACK = True
     OPENSTACK_SDK_IMAGE = openstack.image.v2.image.Image
 except ImportError:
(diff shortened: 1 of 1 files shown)


rebtoor (dismissed):
Change is ok but IMO you should use a meaningful commit message here.

- use some kind of prefix like `[module_utils/image]` in the commit description to help folks identifying what you've changed
- explain the reasons behind the change. I'm seeing a `noqa: S324` so i guess it's due to [this](https://docs.astral.sh/ruff/rules/hashlib-insecure-hash-function/) . It will greatly helps folks that will review the PR that probably don't have the right context.

rebtoor (approved):

PR #773, opened 2026-05-14
Cert checks

fdiazbra:


plugins/modules/_vendor/openstack.cloud
matbu, line 1: I think you added the submodule by mistake here ?

matbu (changes requested):

plugins/modules/_vendor/openstack.cloud
fdiazbra: of course...I think it is one of the generated automatically for the vendor process and I added it to the commit by error at some point

matbu (approved):
Approving the PR and I will fix the end to end job in a follow up.
Since Fernando is away it will be simplier like that.

matbu (approved):
Approving the PR and I will fix the end to end job in a follow up.
Since Fernando is away it will be simplier like that.

PR #793, opened 2026-07-16
Add Makefile target to build container image

clopnis:
This PR adds a new Makefile target that allows building a container
image for os-migrate.

Currently there is no straightforward way to build a containerized
version of os-migrate. This change introduces a `make` target to
simplify building the image locally and in CI/CD pipelines.


cloudguruab (commented):
os-migrate uses containers to run different test cases - https://github.com/os-migrate/os-migrate/blob/main/Makefile#L6

PR #794, opened 2026-07-16
noop commit for running e2e job

cloudguruab:
**Summary**
PR is good to close once bug fix is complete. The action fails (what we expect), the gitlab CI is producing a false-positive after some time. 

Screenshot 2026-07-28 at 5 09 00 PM

CI logs:
```
Could not find versioned identity endpoints when attempting to authenticate. Please check that your auth_url is correct.
...
Task failed: Module failed.
Origin: /root/.ansible/collections/ansible_collections/os_migrate/os_migrate/roles/prelude_src/tasks/main.yml:5:7
3     - os_migrate_src_filter_current_project | default(true) | bool
4   block:
5     - name: Fetch information about currently authenticated user/project
        ^ column 7
<<< caused by >>>
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # OpenStack to Openstack migration tooling
 OS Migrate is an open source toolbox for parallel cloud migration
-between OpenStack clouds.
+between OpenStack clouds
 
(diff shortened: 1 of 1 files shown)


cloudguruab:
Added srcds/dstds to `CLOUDS_FILE` variable set in gitlab settings.

cloudguruab:
Running new job, checking hosts periodically 🤔 seems to be on different steps potentially so not the same error each time

